PR #5, opened 2018-07-28
Use `make install_sw` rather than `make install`

sfackler:
This skips man pages, which take a bunch of time to copy over since there are so many of them.


alexcrichton:
r=me w/ ci

alexcrichton:
nice find!

PR #6, opened 2018-08-03
Add support for emscripten

tomaka:
`wasm32-unknown-unknown` is a bit harder, so I focused on emscripten first.

`emcc` is similar to `gcc`, so we can configure OpenSSL for `gcc` and it works.
However if we simply do that, we get linking errors when building the `openssl` executable because some symbols are defined twice. I have no idea what that happens.

The easiest way to disable building the `openssl` executable if to use `no-stdio`, so that's what I'm doing.

Tests unfortunately fail because of out-of-memory issues, but the ones which don't need a lot of memory all pass.
I didn't add any CI test because it's quite hard to test for emscripten on travis (building emscripten :-/). It should be doable with circle-ci if you think we should add CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: rust
 dist: trusty
-sudo: false
+sudo: required
+
+services:
+  - docker
 
 env:
   global:
@@ -90,6 +93,7 @@ matrix:
         apt:
           packages:
             - gcc-mingw-w64-x86-64
+    - env: TARGET=asmjs-unknown-emscripten CMD=run DOCKER=tomaka/rustc-emscripten
 
     - rust: beta
       env: TARGET=x86_64-unknown-linux-gnu NO_ADD=1 CMD=run
@@ -100,7 +104,11 @@ install:
   - if [ -z "$NO_ADD" ]; then rustup target add $TARGET; fi
 
 script:
-  - cargo "${CMD:-build}" --manifest-path testcrate/Cargo.toml --target $TARGET --bin test
+  - if [ -z "$DOCKER" ]; then
+      cargo "${CMD:-build}" --manifest-path testcrate/Cargo.toml --target $TARGET --bin test
+    else
+      docker run --rm -it -v `pwd`:/usr/code -w /usr/code $DOCKER cargo "${CMD:-build}" --manifest-path testcrate/Cargo.toml --target $TARGET --bin test
+    fi
 
 notifications:
   email:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -124,13 +124,6 @@ impl Build {
             configure.arg("no-shared");
         }
 
-        if target.contains("emscripten") || target == "wasm32-unknown-unknown" {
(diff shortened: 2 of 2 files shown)


tomaka:
I tried bypassing the OOM issues I see, and it turns out all the I/O-related tests fail as we disabled `stdio`.
I think this is not acceptable unfortunately, because this includes a lot of functionalities.

alexcrichton:
Thanks for this! Do you have an Emscripten + OpenSSL build working locally? If so, are they being built slightly differently for one to work where the other isn't?

Additionally, could CI be enabled on Travis via a prebuilt emscripten docker image?

tomaka:
> Do you have an Emscripten + OpenSSL build working locally?

Yes!

> If so, are they being built slightly differently for one to work where the other isn't?

If you're talking about `wasm32-unknown-unknown`, the problem is that `libc` is empty and therefore `rust-openssl` doesn't build. However the compilation of OpenSSL itself for `wasm32-unknown-unknown` works just fine. I added it to this PR.

> Additionally, could CI be enabled on Travis via a prebuilt emscripten docker image?

I'm using this image for all my tests: https://hub.docker.com/r/tomaka/rustc-emscripten/
But is it possible to use docker images on travis?

alexcrichton:
Whoa cool! I was thinking mores for the Emscripten targets with a working build, as I doubt `wasm32-unknown-unknown` will have a working build for awhile (as you've seen we need both a *super* new Clang and a `libc`)

It sounds like though your tests are passing on Emscripten with OpenSSL locally, but not passing when built through this crate? Do you know how the OpenSSL build is different?

alexcrichton:
Oh and, yes, Docker can be used on Travis! There's some [small configuration](https://docs.travis-ci.com/user/docker/) and then it'll have a `docker` command as usual

tomaka:
> It sounds like though your tests are passing on Emscripten with OpenSSL locally, but not passing when built through this crate? Do you know how the OpenSSL build is different?

Ah sorry, I misunderstood you.
The only build I have is through the Rust crate, where many tests fail because of I/O errors.
For some reason if I try to build any OpenSSL binary through OpenSSL's build process, I get duplicate symbols errors.

alexcrichton:
>  The only build I have is through the Rust crate, where many tests fail because of I/O errors.

By this you mean that because of `no-io` that causes `cargo test` in the `openssl` crate itself to fail for asm.js? (I can see it fail locally for Linux)

>  For some reason if I try to build any OpenSSL binary through OpenSSL's build process, I get duplicate symbols errors.

Oh that's totally fine yeah! Having OpenSSL at all on Emscripten seems better than not at all :)

---

I think this should be good to go with a Travis builder for Emscripten (just the asmjs target is fine), that way we can be sure to avoid regressing it with future OpenSSL upgrades (hopefully!)

alexcrichton:
Ok I've tidied up the travis build a bit to go ahead and use docker everywhere and should be good to go, thanks!

PR #7, opened 2018-08-03
Enable parallel builds of OpenSSL on Unix

alexcrichton:
Share Cargo's jobserver with `make` so they can all build in parallel from the
same job pool.


alexcrichton:
@sfackler I think this was fixed in 1.1.0, right?

sfackler:
Yep, parallel builds work in 1.1.0.

PR #10, opened 2018-08-14
Update to cc crate.

ehuss:


ehuss:
I was looking at updating Cargo in the rust repo, and recent changes cause this to add the `gcc` dependency, which I think can be easily avoided.

alexcrichton:
Oh wow forgot about this!

PR #12, opened 2018-09-11
Bump openssl submodule to 1.1.1

alexcrichton:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "openssl-src"
-version = "110.0.7+1.1.0i"
+version = "110.0.8+1.1.1"
```
sfackler: Should this be `110.0.8` or `111.0.0`? Keeping it in the 110 series does mean we don't have to worry about maintaining 1.1.0 and 1.1.1 releases at the same time until 1.1.0 EOLs.

alexcrichton: Er oops yeah, let's switch to `111.0.0`

alexcrichton:
Looks like a ton of build errors across android, musl, and emscripten :(

alexcrichton:
That should fix pc-windows-gnu and Emscripten. I think we need to just delete `linux/mman.h` include to fix musl. For Android it looks like the build just flat out needs more information like the NDK root which we don't currently have (nor do I know how to get in a zero-config fashion). I have no idea what it needs all these tools for.

alexcrichton:
I've hacked up everything but musl at this point, and it seems like the "best" way to fix that will be to basically have a dummy empty header file `linux/mman.h` in some directory we create (and then pass `-I` to that). Turns out musl doesn't actually need the header file to compile!

The hoops we're jumping through to build OpenSSL are a bit ridiculous...

sfackler:
It seems like that may be worth filing upstream.

sfackler:
A workaround mentioned in that thread is to pass `-DOPENSSL_NO_SECURE_MEMORY` to Configure.

alexcrichton:
That sounds like a sort of scary option to pass (in terms of the context here), but perhaps it's good enough for musl?

sfackler:
Yeah it'd definitely not great, but the secure memory feature is new in 1.1.1 so you're not any worse off than using 1.1.0.

alexcrichton:
It does appear to do the trick though!

alexcrichton:
ping @sfackler, do you think we should merge this now that it's green?

PR #14, opened 2018-09-13
Pass no-stdio to OpenSSL's configure

### LINKED ISSUE #13: --- stderr: ***-android-gcc: fatal error: no input files
Hello

I use hyper with hyper-tls, that use native-tls on ios and android.
On Android, native-tls use openssl-sys, and with the feature: vendored, this crate import openssl-src-rs
my cargo.toml file import native-tls like this. native-tls = { version = "0.2", features = ["vendored"] }

So when I compile on this target:
- i686-linux-android
- arm-linux-androideabi
- aarch64-linux-android

I always have the same error:

```
--- stderr
i686-linux-android-gcc: fatal error: no input files
compilation terminated.
i686-linux-android-ar: creating libssl.a
i686-linux-android-ar: creating libcrypto.a
./libcrypto.a(cryptlib.o):cryptlib.c:function OPENSSL_showfatal: error: undefined reference to 'stderr'
collect2: error: ld returned 1 exit status
./libcrypto.a(cryptlib.o):cryptlib.c:function OPENSSL_showfatal: error: undefined reference to 'stderr'
...
...
```

To avoid this problem, in the crate openssl-src-rs, in the file lib.rs, I add 'no-stdio' option to the perl command: configure.

I have just two questions:
Why don't you need to add this option ? and Is it possible to add this modification to the next release ?

Thanks for considering my request,

alexcrichton:
Apparently it doesn't link on Android otherwise and we don't use any of the
functionality anyway!

Closes #13


alexcrichton:
ping @sfackler, any objections to merging?

sfackler:
Do you know why the linker failure wouldn't show up in the CI? It targets android right?

Might be worth rebasing this over the 1.1.1 bump and seeing if that fixes the windows build.

alexcrichton:
We test that android targets build yeah but we don't do a great job of testing it all links currently, we only pull in the `OpenSSL_version_num` symbol and test that links right

alexcrichton:
Alas the answer looks to be [no](https://ci.appveyor.com/project/alexcrichton/openssl-src-rs/build/1.0.95/job/k1b650xoimbbagq2)

sfackler:
Oh well. This seems fine then.

PR #15, opened 2018-09-28
add armv7-linux-androideabi to target list

sercand:
Rust supports 5 different android target and 4 other are in the list except armv7-linux-androideabi.


alexcrichton:
Thanks!

PR #24, opened 2019-04-01
cross compilation on windows for mips

Jedzia:
Just in my case, this can ease my life:)

Install paths are generated that are windows escaped (C:\\for\\the\\way)
When cross compiling to my **mips** target i have GNU-tools working and c:/for/the/way is the way to go.

Use case of Doom: 
Cross compilation for an embedded target and to have the ability to switch between targets. 
Main development happens on the HOST and also testing, playground, debugging.

My solution so far is: 
cfg openssl-sys for linux vendor featured, patch crates-io and have this path local.

```
[target.'cfg(target_os = "linux")'.dependencies]
openssl-sys = { version = "0.9.12", features = ["vendored"] }

[patch.crates-io]
openssl-src = { path = './dependencies/openssl-src-111.2.1+1.1.1b' }

```
full on IDE:
When you have to fiddle with configuration files between switching target/host compilation ... that's annoying. And not our fault. Maybe cargo/xargo shows an easy way in the future for this.

mhghmhmm, Alex: The build triplet 'mipsel-unknown-linux-gnu' is my local egoistic concern and maybe my if clauses are wrong, boss:)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,7 +81,8 @@ impl Build {
         configure.arg("./Configure");
         if target.contains("pc-windows-gnu") {
             configure.arg(&format!("--prefix={}", sanitize_sh(&install_dir)));
-        } else if target.contains("mipsel-unknown-linux-gnu") {
+        } else if host.contains("pc-windows-gnu")
+            && target.starts_with("mips") && target.contains("gnu") {
             configure.arg(&format!("--prefix={}", sanitize_sh(&install_dir)));
         } else {
             configure.arg(&format!("--prefix={}", install_dir.display()));
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,6 +81,8 @@ impl Build {
         configure.arg("./Configure");
         if target.contains("pc-windows-gnu") {
             configure.arg(&format!("--prefix={}", sanitize_sh(&install_dir)));
+        } else if target.contains("mipsel-unknown-linux-gnu") {
+            configure.arg(&format!("--prefix={}", sanitize_sh(&install_dir)));
         } else {
             configure.arg(&format!("--prefix={}", install_dir.display()));
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,10 +79,7 @@ impl Build {
 
         let mut configure = Command::new("perl");
         configure.arg("./Configure");
-        if target.contains("pc-windows-gnu") {
-            configure.arg(&format!("--prefix={}", sanitize_sh(&install_dir)));
-        } else if host.contains("pc-windows-gnu")
-            && target.starts_with("mips") && target.contains("gnu") {
+        if host.contains("pc-windows-gnu") {
             configure.arg(&format!("--prefix={}", sanitize_sh(&install_dir)));
         } else {
             configure.arg(&format!("--prefix={}", install_dir.display()));


alexcrichton:
Thanks for the PR! I think though instead of special-casing mips here, the call to `sanitize_sh` should always happen on a Windows host?

Jedzia:
Good question:) [When did Windows start accepting forward slash as a path separator?](https://bytes.com/topic/python/answers/23123-when-did-windows-start-accepting-forward-slash-path-separator)
It seems that the C:/forward/slash notation is accepted since DOS2.0. The standard notation is of cause C:\backward\slash, but i've seen GNU-Tools with problems with the latter and working happy with forward slash. 
In my case OpenSSL gets the paths for cp wrong. I think they are back-slash notated but not quoted whats leads to "C:ProjectsFolderOtherFolderNextFolder...." paths.

So a target of *-*-linux-gnu cross compiling for on windows will match this case. And according to bytes.com only DOS 1.0 users will be angry ... if they somehow get rustc. 
From OpenSSL NOTES.DJGPP
> If you wish to compile on native DOS with 8+3 filenames, you will
>  have to tweak the installation yourself, including renaming files
>  with illegal or duplicate names.

alexcrichton:
Note that this function isn't about forward vs backwards slash but rather converting a windows path to an MSYS path which looks like `/c/home/...`.

Jedzia:
You are right, That is a MSYS/MinGW problem. So we have to distinguish between x86_64-pc-windows-*msvc* and x86_64-pc-windows-*gnu* like targets when compiled on windows? Or is this a complete separate and extra feature to introduce?
I don't want to break anything or to introduce unnecessary complexity:)

Jedzia:
Here are the detailed working/failing logs for my case.

### unpatched, failing: ###
```
running "perl" "./Configure" "--prefix=E:\\Projects\\Rust\\vuduo2\\vustreamproxy\\target\\mipsel-unknown-linux-gnu\\debug\\build\\openssl-sys-b9ae2d795cc4a957\\out\\openssl-build\\install" "no-dso" "no-ssl3" "no-unit-test" "no-comp" "no-zlib" "no-zlib-dynamic" "no-engine" "no-async" "no-shared" "linux-mips32" "-O2" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-EL"
Configuring OpenSSL version 1.1.1b (0x1010102fL) for linux-mips32
```
which fails with:
`cp: cannot create regular file 'E:ProjectsRustvuduo2vustreamproxytargetmipsel-unknown-linux-gnudebugbuildopenssl-sys-b9ae2d795cc4a957outopenssl-buildinstall/bin/openssl.new': No such file or directory`


### with sanitize_sh(&install_dir), working ###
```
running "perl" "./Configure" "--prefix=/E/Projects/Rust/vuduo2/vustreamproxy/target/mipsel-unknown-linux-gnu/debug/build/openssl-sys-d068b5ef7cd99a7f/out/openssl-build/install" "no-dso" "no-ssl3" "no-unit-test" "no-comp" "no-zlib" "no-zlib-dynamic" "no-engine" "no-async" "no-shared" "linux-mips32" "-O2" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-EL"
Configuring OpenSSL version 1.1.1b (0x1010102fL) for linux-mips32
```
Toolchain:
```
mips-linux-gnu-gcc -v
Configured with: 
...
--build=i686-pc-linux-gnu --host=i686-mingw32
...

```
And yes, this is MSYS/MinGW path formatting ... don't know why i confused this with https://github.com/alexcrichton/openssl-src-rs/pull/24#issuecomment-479782143 (maybe i got infused with stackoverflow laziness?:))

Jedzia:
What do you think about 

```
        } else if host.contains("pc-windows-gnu")
            && target.starts_with("mips") && target.contains("gnu") {
            configure.arg(&format!("--prefix={}", sanitize_sh(&install_dir)));

```
* host.contains("pc-windows-gnu") Identifies the windows host with
* target.starts_with("mips") && target.contains("gnu") that it is a MinGW based build environment 

After accidently closing this PR I had the chance watching the CI work and took a look at https://docs.rust-embedded.org/embedonomicon/compiler-support.html.
There are gnu(abi) and musl toolchains for mips that i have no access to, but according to https://musl.cc/ they are MinGW based and the other stuff is MinGW gcc, too.

Btw. Motivation behind this: 
This opens a way of fluid development for a lot of set-top boxes that are mips based (Enigma2, etc.). Server based applications, IoT, MQTT, etc. with a modern language.
You still have to use a bloated buildroot/OpenEmbedded environment when you build a complete system image. But for a dirty little project Rust is ideal:
Setup toolchain, grab openssl+hyper+reqwest, specify target -> done. And for noobs like me: switch target to host, debug -> insight:)

Jedzia:
i've pushed the changes that worked for me.
and only if host is = pc-windows-gnu we are sanitized

I don't want to specialize in MIPS, but that's the only experience i can tell from. Foresight is your job, chief. ... kidding:) 

_______
For others landing here because they search a solution to RUST-compile OpenSSL for MIPS with gcc/MinGW under Windows :
Don't do what we are discussing here. You are better off, when using the openssl-dev package of your target or develop under linux.

Cargo.toml
`openssl = { version = "0.10", features = ["vendored"] }`

**Is not your way**

`openssl = { version = "x.y" }`

and specifying the environment variable **OPENSSL_DIR** to your development files.

alexcrichton:
Thanks! I think we can safely remove the first branch though (the previous one calling `sanitize_sh` as well as the clause for mips?

alexcrichton:
Thanks!

PR #29, opened 2019-04-11
canonicalize src and dst in cp_r

### LINKED ISSUE #25: canonicalize file names so they can be found on windows
```
C:\build\net\target\wasm32-unknown-unknown\release\build\openssl-sys-965b815f479a83db\out\openssl-build\build\src\pyca-cryptography\vectors\cryptography_vectors\x509\PKITS_data\certpairs\requireExplicitPolicy7subsubsubCARE2RE4CertforwardcrossCertificatePair.cp
```

this has 260 characters, which blows up the `MAX_PATH` on windows when the null terminator is added

canonicalizing file paths avoids this issue to allow the extended length path

https://doc.rust-lang.org/std/fs/fn.canonicalize.html

thedavidmeister:
fixes #25

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -327,6 +327,8 @@ Error {}:
 }
 
 fn cp_r(src: &Path, dst: &Path) {
+    let src = &src.canonicalize().unwrap();
+    let dst = &dst.canonicalize().unwrap();
     for f in fs::read_dir(src).unwrap() {
         let f = f.unwrap();
         let path = f.path();


alexcrichton:
Thanks! Since `canonicalize` is only really requires as a best-effort feature on Windows could this be documented as such and could the error case be handled by using `unwrap_or`? Additionally I think that this probably only needs to canonicalize once at the top, right?

thedavidmeister:
@alexcrichton the unwraps are extremely painful for debugging FYI because all you get is an opaque "specified path" type error with no contextual information about what went wrong

i thought that my PR does only do this once at the top, where are you thinking it should be moved to?

what i found with `canonicalize` in my testing, and from the docs, is that it uses the extended path length:

https://docs.microsoft.com/en-gb/windows/desktop/FileIO/naming-a-file#maximum-path-length-limitation

which looks like `\\?\` as a prefix

if we don't want to canonicalize we can manually modify the prefix using strings or similar (which i also tested, and works) but that didn't seem like the "right way" to fix this, what do you think?

alexcrichton:
This function is recursive so it's canonicalizing on each recursive call, but only the first call at the very beginning needs to have canonicalized paths.

thedavidmeister:
@alexcrichton ah ok, would you be able to update the code to reflect that?

alexcrichton:
Sorry but I don't personally have the ability to test this right now, and I wouldn't want to land this without verifying the fix.

thedavidmeister:
@alexcrichton but you won't accept the fix we have now, imperfect though it may be?

alexcrichton:
No, I would prefer that known bugs were fixed before merging.

thedavidmeister:
@alexcrichton sorry, i'm unsure what is a bug?

i expect canonical paths to be harmless even if recursive

alexcrichton:
@thedavidmeister I'm sorry but I had comments on this PR about what I would like changed before merging. If you're not up to making those changes at this time that's ok and this can wait until there's time and/or energy. I do not personally have the time/energy to make the changes. I will not merge the PR until the changes have been applied because I asked for changes.

thedavidmeister:
"because i asked for changes"... haha ok, good reason :ok_hand:

articulate what is bad in recursive canonicalisation? with a reason i will update
i have energy for improvements, not arbitrary LOC shuffling

i think you should merge "because i found and fixed a real bug that impacts real downstream users" :beetle:

alexcrichton:
Sorry but it sounds clear like you're not willing to make more changes here, and I'd still like to see changes before this merges but don't have time to do them myself, so I'm going to close this.

thedavidmeister:
@alexcrichton sorry but that's a shitty response

i fixed a bug and you can't explain _why anyone_ should make the change you requested

i'm absolutely 100% certainly willing to make changes if they make a material difference

alexcrichton:
@thedavidmeister I'm sorry if you feel that way but I'd ask you to refrain from calling things "shitty" when responding. This indeed fixes a bug and I [made a comment](https://github.com/alexcrichton/openssl-src-rs/pull/29#issuecomment-482138978) explaining a change I would like to make. The change is necessary to avoid unnecessary calls to `canonicalize` which can slow things down and unnecessarily introduce error conditions.

If you're not willing to make any changes that's ok, but I'd rather not continue to drag on this PR for months after I asked for changes with repeated requests to merge this when such changes have not happened.

thedavidmeister:
@alexcrichton are they definitely unnecessary though?

windows is going to error any time the path doesn't follow the canonical format if the file name is too long

i would have expected this means we need to be handling everything recursively

the slowdown would be a micro-optimisation at best, it's just making minor tweaks to path strings

:thinking: 

it *is* shitty that you feel you chose to talk on my behalf as "it's clear you're not willing to make changes" and close out the issue without holding space for discussion in good faith

what tz you in, we could pair on zoom or similar and land this together?

i'll make the changes you requested once i better understand them

alexcrichton:
They indeed definitely are unnecessary for the task being solved at hand here, there's no need to canonicalize elsewhere. I'm sorry but this is just an open source project and nowhere near my primary focus. I try to make sure I review any contribution coming in to any project I'm maintaining (which is far too many) and it's quite draining on a contribution like this to have a months-long slog of back and forth. I feel like as a contributor if a maintainer asks for a change it shouldn't result in months of conversation but rather some investigation and research if it's understood (and clarifications too if needed).

I get the impression you're not willing to make changes because the fix here is effectively a one line change by moving the canonicalization to a different location. It's been months now at this point that you're *very* insistent that this exact patch should be merged.

I'm sorry but I don't have time to jump on a video call for something like this.

PR #32, opened 2019-05-31
Upgrade to OpenSSL 1.1.1c

sfackler:


sfackler:
Looks like this MUSL break will be harder to hack around than the last one :(

https://github.com/openssl/openssl/blob/OpenSSL_1_1_1c/crypto/rand/rand_unix.c#L21-L23

alexcrichton:
r=me with a green build, although I don't know how best to fix musl.

sfackler:
MUSL build is green - does the patch logic look okay to you?

PR #36, opened 2019-08-12
add aarch64-apple-ios target support

arrtchiu:
I'm not sure how to test this, please let me know if you have a tip on how to get started!

Hacks:
- Remove `-arch xxx` arg supplied by `cargo-lipo`; OpenSSL already adds it and complains
- Remove and split up `-isysroot xxx` arg supplied by `cargo-lipo`; OpenSSL already adds it and complains, but needs parts of it during *build* phase


alexcrichton:
👍

PR #38, opened 2019-09-09
Fix cross compilation when `CROSS_COMPILE` env is set

oblique:
When `CROSS_COMPILE` is set `cc::Build` is using it to generate compiler command. However `./Configure` of OpenSSL is using it again which has as a result the environment variable to the prepended twice.

For example `CROSS_COMPILE=arm-none-linux-gnueabi` will result to
`arm-none-linux-gnueabiarm-none-linux-gnueabi-gcc`


alexcrichton:
Thanks!

PR #39, opened 2019-09-11
Update to 1.1.1d

sfackler:


alexcrichton:
Windows ci is broken due to an environment issue, I'll post a fix soon so we can make sure this doesn't accidentally regress there

PR #49, opened 2020-01-09
Allow to specify custom path for Perl

### LINKED ISSUE #45: Allow for the path to the perl installation to be specified
Currently, the crate just runs `perl` which will use the first installation that is in the path.

This this not work on the windows-runner of GitHub actions and fails with the following error:

```
error: failed to run custom build command for `openssl-sys v0.9.53`

Caused by:
  process didn't exit successfully: `d:\a\create-comit-app\create-comit-app\target\debug\build\openssl-sys-5ec2b21008c71a01\build-script-main` (exit code: 101)
--- stdout
cargo:rustc-cfg=const_fn
running "perl" "./Configure" "--prefix=d:\\a\\create-comit-app\\create-comit-app\\target\\debug\\build\\openssl-sys-935999bfd09d3702\\out\\openssl-build\\install" "no-dso" "no-ssl3" "no-unit-test" "no-comp" "no-zlib" "no-zlib-dynamic" "no-engine" "no-async" "no-asm" "VC-WIN64A"
Configuring OpenSSL version 1.1.1d (0x1010104fL) for VC-WIN64A
Using os-specific seed configuration

--- stderr

******************************************************************************
This perl implementation doesn't produce Windows like paths (with backward
slash directory separators).  Please use an implementation that matches your
building platform.

This Perl version: 5.26.2 for x86_64-msys-thread-multi
******************************************************************************
```

Adding the already installed version of strawperryperl to the path fixes the issue.

However, it would be nice to have an environment variable that allows one to override the path to the Perl installation. This would make it more obvious, what we need the different Perl installation for.

I am thinking of an env variable like `OPENSSL_SRC_PERL_HOME`.

Thoughts?

D4nte:
Resolves #45

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -122,7 +122,7 @@ jobs:
       shell: bash
     - name: Use strawberry perl
       if: startsWith(matrix.os, 'windows')
-      run: echo ::add-path::C:/Strawberry/perl/bin
+      run: echo ::set-env name=OPENSSL_SRC_PERL::C:/Strawberry/perl/bin/perl
       shell: bash
     - run: |
         set -e
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,7 +77,8 @@ impl Build {
         cp_r(&source_dir(), &inner_dir);
         apply_patches(target, &inner_dir);
 
-        let perl_program = env::var("OPENSSL_SRC_PERL").unwrap_or(String::from("perl"));
+        let perl_program =
+            env::var("OPENSSL_SRC_PERL").unwrap_or(env::var("PERL").unwrap_or("perl".to_string()));
         let mut configure = Command::new(perl_program);
         configure.arg("./Configure");
         if host.contains("pc-windows-gnu") {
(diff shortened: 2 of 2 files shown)


alexcrichton:
Given what @sfackler mentioned [here](https://github.com/alexcrichton/openssl-src-rs/issues/45#issuecomment-565217393) it seems like we may want to check this env var, but fall back to `PERL` if it's set as well?

D4nte:
> Given what @sfackler mentioned [here](https://github.com/alexcrichton/openssl-src-rs/issues/45#issuecomment-565217393) it seems like we may want to check this env var, but fall back to `PERL` if it's set as well?

Sounds good. I am still trying to make this work in my project. Will clean up this PR once confirmed working as expected and implement the fallback logic.

D4nte:
Not sure why the tests where not passing yesterday but it looks good to go now.

JasonKleban:
Should this work in cargo.toml on Windows, please?

```
...

[env]
OPENSSL_SRC_PERL = "C:/myperl/perl/bin"
```

assuming the perl.exe is in that bin folder?

PR #53, opened 2020-03-17
Use gmake by default on all FreeBSD or DragonFlyBSD targets

MikaelUrankar:
Or I could list all the FreeBSD targets...


alexcrichton:
I think this may have lost the recent adjustment for solaris?

MikaelUrankar:
Indeed, I cloned the repo at the wrong time :)

Issue #61, opened 2020-05-13
AR environment variable is overwritten


jamesmcm:
Hi,

This line overwrites the `AR` environment variable if `gcc` has a prefix.

https://github.com/alexcrichton/openssl-src-rs/blob/babdbf06116b22f8fa276da4e25b134d28b54494/src/lib.rs#L222

This is a pain as my `ar` is  `x86_64-unknown-linux-gnu-gcc-ar` not `x86_64-unknown-linux-gnu-ar` - is there any way to override this?

It's running on a build server so I don't have much control.

alexcrichton:
This is likely a bug and something which we should avoid doing if `AR` is already specified.

jamesmcm:
Yeah, in the end I switched the code to use Rustls and it works (though I had a few issues getting the build system to use Clang to compile ring).

jamesmcm:
Addressed in PR #62

PR #62, opened 2020-05-17
Do not overwrite AR and RANLIB env vars if set

jamesmcm:
Addresses #61 

I wasn't sure of the best way to test this, but it's a small fix in any case.


alexcrichton:
Thanks!

PR #63, opened 2020-06-06
Add engine support for linux-gnu

hunterlxt:
I met link error when updating in grpc-rs which is a binding lib to grpc-core. And I found openssl-src remove `engine` on no matter what platform.

Related PR: https://github.com/tikv/grpc-rs/pull/466

Signed-off-by: Xintao <hunterlxt@live.com>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,13 +107,9 @@ impl Build {
             // stuff depends on, and we don't bind to any of that in any case.
             .arg("no-async");
 
-        if !target.contains("x86_64-unknown-linux-gnu") {
+        if target.contains("musl") {
             // This actually fails to compile on musl (it needs linux/version.h
-            // right now) but we don't actually need this most of the time. This
-            // is intended for super-configurable backends and whatnot
-            // apparently but the whole point of this script is to produce a
-            // "portable" implementation of OpenSSL, so shouldn't be any harm in
-            // turning this off.
+            // right now) but we don't actually need this most of the time.
             configure.arg("no-engine");
         }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -102,15 +102,19 @@ impl Build {
             // Nothing related to zlib please
             .arg("no-comp")
             .arg("no-zlib")
-            .arg("no-zlib-dynamic")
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -103,16 +103,19 @@ impl Build {
             .arg("no-comp")
             .arg("no-zlib")
             .arg("no-zlib-dynamic")
+            // MUSL doesn't implement some of the libc functions that the async
+            // stuff depends on, and we don't bind to any of that in any case.
+            .arg("no-async");
+
+        if !target.contains("x86_64-unknown-linux-gnu") {
```
BusyJay, line 110: I think better check for `musl` specifically so that other OS can also utilize engine.

hunterlxt:
@alexcrichton PTAL thanks

src/lib.rs
```diff
@@ -103,17 +103,16 @@ impl Build {
...
-            // "portable" implementation of OpenSSL, so shouldn't be any harm in
-            // turning this off.
-            .arg("no-engine")
             // MUSL doesn't implement some of the libc functions that the async
             // stuff depends on, and we don't bind to any of that in any case.
             .arg("no-async");
 
+        if target.contains("musl") || target.contains("android") || target.contains("windows") {
+            // This actually fails to compile on musl (it needs linux/version.h
+            // right now) but we don't actually need this most of the time.
```
alexcrichton, line 112: Could this comment be updated as to why this option is skipped for each of the platforms here?

hunterlxt: I tried to make windows pass CI but failed. Comments is added

hunterlxt:
@alexcrichton  We use openssl-sys in our repo. My original plan was to use openssl-src master branch by using patch, and then wait for openssl-src to release a new version on crates.io. 
Unfortunately, because of this [bug](https://github.com/rust-lang/git2-rs/pull/567), cargo cannot pull https://boringssl.googlesource.com/boringssl, and waiting for the repair of the official version of cargo is too long. 
Thanks a lot if you can release a new version on crates.io. So we don't need to fork a copy.

PR #67, opened 2020-07-08
Fix build on macOS with latest cc crate

benesch:
cc v1.0.58 broke the macOS build by including the "-arch" flag in the
default set of compiler flags. Strip it out, like we do for iOS targets.

Closes #66.


alexcrichton:
👍 
Thanks!

PR #68, opened 2020-07-11
add optional features for less used algorithms

petreeftime:
This commit disables by default a few of the weaker cryptographical algorithms into
a "weak-crypto" feature as well as some of the less used algorithms into
their own specific features.

These algorithms are not directly exposed through the rust-openssl crate.
The compilation of these can be re-enabled by selecting the desired features.
This should slightly reduce build time and library size.

Signed-off-by: Petre Eftime <petre.eftime@gmail.com>

## Testing
I've tested this feature by building this crate with cargo, and by modifying a local rust-openssl repo to use my local clone as the source for openssl-src, and ran `cargo test --features vendored`, with and without these features specified in Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,21 @@ exclude = [
   'openssl/test/*',
 ]
 
+[features]
+default = []
+# Enables compilation of some older algorithms: md2 (hash), rc5 (block cypher) and enabled use of
+# some weaker algorithms in SSL connections. These are generally not recommended for use.
+weak-crypto = []
+# Enables compilation of the Camellia symmetric key block cypher. Since hardware acceleration for
+# it is not available on most systems, this is not as used as AES.
+camellia = []
+# Enables compilation of International Data Encryption Algorithm (IDEA), a symmetric key block
+# cypher sometimes used as an AES128 alternative.
+idea = []
+# Enables compilation of SEED, a symmetric key block cypher mostly used in South Korea, but
+# otherwise not widely supported.
+seed = []
+
 [workspace]
 members = ['testcrate']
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -104,6 +104,23 @@ impl Build {
             .arg("no-zlib")
             .arg("no-zlib-dynamic");
 
+        #[cfg(not(feature = "weak-crypto"))]
```
alexcrichton, line 107: Instead of `#[cfg]` could these blocks use `if cfg!(...)`?

Cargo.toml
```diff
@@ -16,6 +16,13 @@ exclude = [
   'openssl/test/*',
 ]
 
+[features]
+default = []
+weak-crypto = []
+camellia = []
+idea = []
+seed = []
```
alexcrichton, line 24: Could you add comments to each of these features indicating what they enable? (I'm not sure what "camellia", "seed", or "idea" are)

alexcrichton:
Could you also add a test on CI to just one builder that `build --all-features` works?

Cargo.toml
petreeftime: Added in Cargo.toml, but maybe these features should be documented in the README as well?

petreeftime:
> Could you also add a test on CI to just one builder that `build --all-features` works?

I initially tried adding an extra step in the workflow for stable, but I couldn't find an easy way to run this in the container, so I've added this to the `ci/run.sh` file instead, but I'm not convinced of this approach. I was considering adding a second parameter to `ci/run-docker.sh` that would allow setting a different script to run instead. Would this be better?

alexcrichton:
Looks good to me, thanks!

PR #74, opened 2020-09-26
Add upstream patch to allow building on aarch64-apple-darwin

### LINKED ISSUE #72: Supporting aarch64-apple-darwin
I was hoping that OpenSSL would accept [the PR that adds support to their build system for aarch64-apple-darwin](https://github.com/openssl/openssl/pull/12369), but with 1.1.1h released without it, it seems unlikely by the end of 2020. As such, I'd like to discuss how the project feels about **applying our own patches** on top of the official release. 

Specifically, the patch in question only modifies the build system to use existing pieces in new ways.

With that applied, and a single line added to this crate to map target triples, it's possible to cross-compile to aarch64-apple-darwin:

```
% SDKROOT=$(xcrun -sdk macosx11.0 --show-sdk-path) MACOSX_DEPLOYMENT_TARGET=$(xcrun -sdk macosx11.0 --show-sdk-platform-version) cargo +462ec6baef52edea591eab1a070ed4e18b80d094 build --target=aarch64-apple-darwin --features vendored
   Compiling cc v1.0.60
   Compiling pkg-config v0.3.18
   Compiling autocfg v1.0.1
   Compiling libc v0.2.77
   Compiling bitflags v1.2.1
   Compiling foreign-types-shared v0.1.1
   Compiling openssl v0.10.30 (/Users/shep/Projects/rust-openssl/openssl)
   Compiling cfg-if v0.1.10
   Compiling lazy_static v1.4.0
   Compiling foreign-types v0.3.2
   Compiling openssl-src v111.10.2+1.1.1g (/Users/shep/Projects/openssl-src-rs)
   Compiling openssl-sys v0.9.58 (/Users/shep/Projects/rust-openssl/openssl-sys)
    Finished dev [unoptimized + debuginfo] target(s) in 29.71s
```

This should make it possible to build Cargo for this platform as well.

shepmaster:
Closes #72

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -157,6 +157,7 @@ impl Build {
         }
 
         let os = match target {
+            "aarch64-apple-darwin" => "darwin64-arm64-cc",
             // Note that this, and all other android targets, aren't using the
             // `android64-aarch64` (or equivalent) builtin target. That
             // apparently has a crazy amount of build logic in OpenSSL 1.1.1
@@ -446,6 +447,11 @@ fn cp_r(src: &Path, dst: &Path) {
 }
 
 fn apply_patches(target: &str, inner: &Path) {
+    apply_patches_musl(target, inner);
+    apply_patches_aarch64_apple_darwin(target, inner);
+}
+
+fn apply_patches_musl(target: &str, inner: &Path) {
     if !target.contains("musl") {
         return;
     }
@@ -458,6 +464,40 @@ fn apply_patches(target: &str, inner: &Path) {
     let buf = buf
         .replace("asm/unistd.h", "sys/syscall.h")
         .replace("__NR_getrandom", "SYS_getrandom");
+
+    fs::write(path, buf).unwrap();
+}
+
+fn apply_patches_aarch64_apple_darwin(target: &str, inner: &Path) {
+    if target != "aarch64-apple-darwin" {
+        return;
+    }
+
+    // Apply build system changes to allow configuring and building
+    // for Apple's ARM64 platform.
+    // https://github.com/openssl/openssl/pull/12369
+
+    let path = inner.join("Configurations/10-main.conf");
+    let mut buf = fs::read_to_string(&path).unwrap();
+
+    assert!(
+        !buf.contains("darwin64-arm64-cc"),
+        "{} already contains instructions for aarch64-apple-darwin",
+        path.display(),
+    );
+
+    const PATCH: &str = r#"
+    "darwin64-arm64-cc" => {
+        inherit_from     => [ "darwin-common", asm("aarch64_asm") ],
+        CFLAGS           => add("-Wall"),
+        cflags           => add("-arch arm64"),
+        lib_cppflags     => add("-DL_ENDIAN"),
+        bn_ops           => "SIXTY_FOUR_BIT_LONG",
+        perlasm_scheme   => "ios64",
+    },"#;
+
+    let x86_64_stanza = buf.find(r#"    "darwin64-x86_64-cc""#).unwrap();
+    buf.insert_str(x86_64_stanza, PATCH);
 
     fs::write(path, buf).unwrap();
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,7 @@
 extern crate cc;
 
 use std::env;
-use std::fs::{self, File};
-use std::io::{Read, Write};
+use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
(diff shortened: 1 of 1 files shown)


shepmaster:
With this branch, I was able to cross-compile Cargo:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index aa6be2f3d..fb8986e0c 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,3 +119,6 @@ doc = false
 deny-warnings = []
 vendored-openssl = ["openssl/vendored"]
 pretty-env-logger = ["pretty_env_logger"]
+
+[patch.crates-io]
+openssl-src = { git = "https://github.com/shepmaster/openssl-src-rs",  branch = "aarch64-apple-darwin" }
```

```
% SDKROOT=$(xcrun -sdk macosx11.0 --show-sdk-path) MACOSX_DEPLOYMENT_TARGET=$(xcrun -sdk macosx11.0 --show-sdk-platform-version) cargo +462ec6baef52edea591eab1a070ed4e18b80d094 build --target=aarch64-apple-darwin --features vendored-openssl

    Finished dev [unoptimized + debuginfo] target(s) in 1m 34s
```

alexcrichton:
Is it possible to test this on CI? Or do the macs on github actions not have the sdk installed?

shepmaster:
> Is it possible to test this on CI

It’s *kind* of a catch 22 situation. This change is needed to be able to build Cargo for aarch64 in Rust CI. 

Right now, the only artifact you could use would be a random try build. With that, and some environment variables, you could cross compile to aarch64 on GitHub Actions or Azure Pipelines (haven’t looked at other CI providers). 

None of the CI providers have aarch64 Darwin hardware, so you can’t actually run any tests.

shepmaster:
Once https://github.com/rust-lang/rust/pull/75991 is merged, then presumably those artifacts would be available to rustup in nightly, making it one step easier.

src/lib.rs
```diff
@@ -467,6 +473,44 @@ fn apply_patches(target: &str, inner: &Path) {
...
+        cflags           => add("-arch arm64"),
+        lib_cppflags     => add("-DL_ENDIAN"),
+        bn_ops           => "SIXTY_FOUR_BIT_LONG",
+        perlasm_scheme   => "ios64",
+    },"#;
+
+    let x86_64_stanza = buf.find(r#"    "darwin64-x86_64-cc""#).unwrap();
+    buf.insert_str(x86_64_stanza, PATCH);
+
+    File::create(&path)
```
sfackler, line 508: This can just use `fs::write(&path, buf.as_bytes())`

src/lib.rs
```diff
@@ -467,6 +473,44 @@ fn apply_patches(target: &str, inner: &Path) {
...
+        return;
+    }
+
+    // Apply build system changes to allow configuring and building
+    // for Apple's ARM64 platform.
+    // https://github.com/openssl/openssl/pull/12369
+
+    let mut buf = String::new();
+    let path = inner.join("Configurations/10-main.conf");
+    File::open(&path).unwrap().read_to_string(&mut buf).unwrap();
```
sfackler, line 487: This can use `fs::read_to_string`

src/lib.rs
shepmaster: Ok, I just copied the existing patch structure and assumed that the library was striving to support older Rust versions. 

Would you like me to clean up the MUSL to use these functions as well?

sfackler: The oldest version we need to support is 1.31 (at least that's what rust-openssl tests against), so the fs:: methods should be fine there. Yeah, might as well clean up the MUSL one as well.

src/lib.rs
```diff
@@ -467,6 +473,44 @@ fn apply_patches(target: &str, inner: &Path) {
...
+
+    assert!(
+        !buf.contains("darwin64-arm64-cc"),
+        "{} already contains instructions for aarch64-apple-darwin",
+        path.display(),
+    );
+
+    const PATCH: &'static str = r#"
```
est31, line 495: `'static` has not been required on constants [since 1.17.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#language-28)

shepmaster: Yup. Again, just being overly cautious because I wasn't aware of the crate's MSRV requirements.

alexcrichton:
Ah right sorry forgot that we needed a Rust target. In that case this looks good to me, thanks again @shepmaster!

shepmaster:
I’ll try to submit a CI PR once the target lands in nightly. I assume that testing via cross compiling and with nightly should be ok?

shepmaster:
Thanks!

alexcrichton:
Yeah that'd be great! And yeah I don't expect to run tests, but it would be good to exercise the patching and building logic if we can.

PR #76, opened 2020-09-30
i586 support

mon:
I'm working on an Intel Quark which actually uses such an old instruction set.

Adding the `i586-unknown-linux-musl` target was all that was required to get a successful TLS connection with reqwest, compiled using the i586 toolchain provided by rust-embedded/cross.

I haven't explicitly tested the `i586-unknown-linux-gnu` target but I have no reason to think it won't work, as musl is usually the trickier target.


alexcrichton:
Thanks!

PR #81, opened 2020-12-26
Make it compile for wasm32-wasi

tomaka:
It seems that `wasm32-unknown-unknown` and `wasm32-unknown-emscripten` are already supported. This PR adds `wasm32-wasi` to this list.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -218,6 +218,7 @@ impl Build {
             "x86_64-sun-solaris" => "solaris64-x86_64-gcc",
             "wasm32-unknown-emscripten" => "gcc",
             "wasm32-unknown-unknown" => "gcc",
+            "wasm32-wasi" => "gcc",
             "aarch64-apple-ios" => "ios64-cross",
             "x86_64-apple-ios" => "iossimulator-xcrun",
             _ => panic!("don't know how to configure OpenSSL for {}", target),


tomaka:
Note that I had to apply this patch to OpenSSL for the crate to actually compile:

```diff
diff --git a/apps/openssl.c b/apps/openssl.c
index 2607694f59..608977693d 100644
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -24,7 +24,7 @@
 #ifdef OPENSSL_FIPS
 # include <openssl/fips.h>
 #endif
-#define USE_SOCKETS /* needed for the _O_BINARY defs in the MS world */
+//#define USE_SOCKETS /* needed for the _O_BINARY defs in the MS world */
 #include "s_apps.h"
 /* Needed to get the other O_xxx flags. */
 #ifdef OPENSSL_SYS_VMS
```

alexcrichton:
Seems like a reasonable start at least!

FrankC01:
@tomaka  - Can you be more specific about what we need to do to compile for `wasm32-wasi` ? Not sure what changes where made to signal.h, etc.

tomaka:
For signal.h, it's [pretty self-explanatory](https://github.com/WebAssembly/wasi-libc/blob/c82d116de0de47bf2312c81aa536d22cb0865f5b/libc-top-half/musl/include/signal.h#L2-L3). I removed the `#ifndef` guard so that I didn't have to pass any flag.

I don't remember, and unfortunately my local copy is clean, but I believe that the change that I had to do on OpenSSL itself was related to `pthread`. If my memory is correct, there was at the time a PR open in `wasi-libc` to fix that, and since didn't want to bother too much I just tweaked OpenSSL.

Issue #59, opened 2020-04-08
nasm support for msvc?


KaneGreen:
I've noticed [this code](https://github.com/alexcrichton/openssl-src-rs/blob/master/src/lib.rs#L125-L127) in the `lib.rs`, which indicates that `no-asm` is enabled for msvc toolchain no matter `nasm.exe` is installed or not.

As far as I know, assembly language routines of OpenSSL provide hardware acceleration support such as AES-NI, which is a performance booster.

Installing `nasm.exe` is easy. So, I think we should consider the users who installed it and disable `no-asm` for them.

You can consider this code:
```Rust
use std::process::Command;
let wherenasm = Command::new("cmd")
    .args(&["/C", "where nasm"])
    .output()
    .expect("failed to exec cmd");
if !wherenasm.status.success() {
    configure.arg("no-asm");
}
```
This modification works on Windows 10 64bit (1909), VS Community 2019 (16.5.2), rustc 1.42.0 (stable-x86_64-pc-windows-msvc), NASM version 2.14.02.
```TOML
[dependencies]
openssl = { version = "0.10", features = ["vendored"] }
[patch.crates-io]
openssl-src = { path = '../PatchedCrates/openssl-src-111.8.1+1.1.1f' }
```

sfackler:
We could also potentially vendor the nasm binary - it looks like it's only ~500kb zipped and licensed BSD 2-clause.

KaneGreen:
> We could also potentially vendor the nasm binary - it looks like it's only ~500kb zipped and licensed BSD 2-clause.

There might be someone who wanna compile by himself/herself. I think providing a way is not a bad idea.

KaneGreen:
PR merged. So closed.

PR #87, opened 2021-03-16
add nasm support for msvc

KaneGreen:
issue #59

For Windows users, if the `OPENSSL_RUST_USE_NASM` environment variable is set, the assembly language routines will be enabled.

Tested on Windows 10 64bit (20H2), VS Community 2019 (16.9.1), rustc 1.50.0 (stable-x86_64-pc-windows-msvc), NASM version 2.15.05.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -107,6 +107,7 @@ jobs:
         curl -O https://www.nasm.us/pub/nasm/releasebuilds/${WINNASMVERSION}/win64/nasm-${WINNASMVERSION}-win64.zip
         unzip nasm-${WINNASMVERSION}-win64.zip
         echo "$PWD/nasm-${WINNASMVERSION}" >> $GITHUB_PATH
+        echo "OPENSSL_RUST_USE_NASM=yes" >> $GITHUB_ENV
     - name: Run tests (Windows)
       if: startsWith(matrix.os, 'windows')
       run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -114,13 +114,16 @@ jobs:
         WINNASMVERSION='2.15.05'
         curl -O https://www.nasm.us/pub/nasm/releasebuilds/${WINNASMVERSION}/win64/nasm-${WINNASMVERSION}-win64.zip
         unzip nasm-${WINNASMVERSION}-win64.zip
-        echo PATH=$PWD/nasm-${WINNASMVERSION}:$PATH >> $GITHUB_ENV
+        mv nasm-${WINNASMVERSION} nasm_win64
     - name: Set NO_NASM env var (Windows)
       if: startsWith(matrix.os, 'windows') && matrix.env_no_nasm == 'set'
       run: echo OPENSSL_RUST_NO_NASM=1 >> $GITHUB_ENV
     - name: Run tests (Windows)
       if: startsWith(matrix.os, 'windows')
       run: |
+        if [ '${{ matrix.nasm_exe }}' = 'installed' ] ; then
+        export PATH=$PWD/nasm_win64:$PATH
+        fi
         cargo test --manifest-path testcrate/Cargo.toml --target ${{ matrix.target }} -vv
         cargo test --manifest-path testcrate/Cargo.toml --target ${{ matrix.target }} --release -vv
         cargo run --release --target ${{ matrix.target }} --manifest-path testcrate/Cargo.toml --features package -vv
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,30 +109,21 @@ jobs:
         cargo generate-lockfile
         ./ci/run-docker.sh ${{ matrix.target }}
     - name: Download nasm.exe (Windows)
-      if: startsWith(matrix.os, 'windows') && matrix.nasm_exe == 'installed'
+      if: matrix.nasm_exe == 'installed'
       run: |
         WINNASMVERSION='2.15.05'
         curl -O https://www.nasm.us/pub/nasm/releasebuilds/${WINNASMVERSION}/win64/nasm-${WINNASMVERSION}-win64.zip
         unzip nasm-${WINNASMVERSION}-win64.zip
-        mv nasm-${WINNASMVERSION} nasm_win64
+        echo "$PWD/nasm-${WINNASMVERSION}" >> $GITHUB_PATH
     - name: Set NO_NASM env var (Windows)
-      if: startsWith(matrix.os, 'windows') && matrix.env_no_nasm == 'set'
+      if: matrix.env_no_nasm == 'set'
       run: echo OPENSSL_RUST_NO_NASM=1 >> $GITHUB_ENV
     - name: Run tests (Windows)
       if: startsWith(matrix.os, 'windows')
       run: |
-        if [ '${{ matrix.nasm_exe }}' = 'installed' ] ; then
-        export PATH=$PWD/nasm_win64:$PATH
-        fi
         cargo test --manifest-path testcrate/Cargo.toml --target ${{ matrix.target }} -vv
         cargo test --manifest-path testcrate/Cargo.toml --target ${{ matrix.target }} --release -vv
         cargo run --release --target ${{ matrix.target }} --manifest-path testcrate/Cargo.toml --features package -vv
-    - name: Verify no-nasm option (Windows)
-      if: startsWith(matrix.os, 'windows') && matrix.env_no_nasm == 'set'
-      run: |
-        # if the nasm.exe is disabled, there should be 'no-asm' in the configure command
-        echo 'CI: In output files, there should be no-asm configured:'
-        grep 'no-asm' target/${{ matrix.target }}/*/build/*/output
 
   rustfmt:
     name: Rustfmt
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -107,7 +107,7 @@ jobs:
         curl -O https://www.nasm.us/pub/nasm/releasebuilds/${WINNASMVERSION}/win64/nasm-${WINNASMVERSION}-win64.zip
         unzip nasm-${WINNASMVERSION}-win64.zip
         echo "$GITHUB_WORKSPACE\\nasm-${WINNASMVERSION}" >> $GITHUB_PATH
-        echo "OPENSSL_RUST_USE_NASM=yes" >> $GITHUB_ENV
+        echo "OPENSSL_RUST_USE_NASM=1" >> $GITHUB_ENV
     - name: Run tests (Windows)
       if: startsWith(matrix.os, 'windows')
       run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,9 @@ defaults:
   run:
     shell: bash
 
+env:
+  WINNASMVERSION: 2.15.05
+
 jobs:
   test:
     name: Test
@@ -76,6 +79,15 @@ jobs:
           rust: stable-x86_64-msvc
           os: windows-latest
           crt_static: yes
+        - target: x86_64-pc-windows-msvc
+          rust: stable-x86_64-msvc
+          os: windows-latest
+          nasm_exe: installed
+        - target: x86_64-pc-windows-msvc
+          rust: stable-x86_64-msvc
+          os: windows-latest
+          nasm_exe: installed
+          env_no_nasm: set
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Can this be tested on  CI? (both with and without nasm ideally)

Additionally I think the logic is inverted? It looks like if the env var is unset then this panics? I think it's fine to have the env var to forcibly disable or forcibly enable, but otherwise detecting `nasm` at build time seems like the best choice.

KaneGreen:
I tested this on my local machine, not a CI.

This logic is a bit confusing. According to my understanding and test results, it works like this:
function `is_nasm_ready()` return value:
|  | env is set | env not set or =0 |
| -- | -- | -- |
| nasm is installed | true | false |
| nasm not installed | panic! | false |

In other words, by default the env var is not set, so nasm is disabled. When the env var is set to a non-zero value, it detects `nasm` (and panic when `nasm` is not installed because this env var conflicts with the absence of `nasm`)

alexcrichton:
Oh oops I never know how to read `map_or`, so I got mixed up.

I still think, though, that maybe an env var isn't worth it here? Could `nasm.exe` just be detected and we enable it whether it's installed or not?

KaneGreen:
New version:
1. the env var changed to `OPENSSL_RUST_NO_NASM`, which disable the asm compiling.
2. the default behavior is auto-detecting.

function `is_nasm_ready()` return value:
|  | env var is set | env var not set or =0 |
| -- | -- | -- |
| nasm is installed | false ❌ | true ✔ |
| nasm not installed | false ❌ | false ❌ |

In other words, by default the env var is not set, `nasm.exe` will be auto-detected. But user can use the `OPENSSL_RUST_NO_NASM` env var to disable it anyway.

-------------
I think the env var is necessary. It is a switch to control whether to use asm compiling, for it is not easy to temporarily [exclude a binary from the PATH](https://unix.stackexchange.com/questions/350905/exclude-single-executable-from-path-without-chmod-x-it).
What will happen if we just detect `nasm.exe` without noticing users? Considering there are two people using build environments with the same configuration expect one installed `nasm.exe` but the other does not, they use the same command to build the same code -- but they get two different binaries (I am not referring to the difference in very details, but to the significant difference caused by the difference in the code/objects).
So, I think the env var is worth.

alexcrichton:
Looks good to me! Can you ensure that this is tested on CI?

KaneGreen:
CI Test added. Now, the windows test on GitHub Actions will takes about 25 min.
On my fork, it passed. And on my local windows PC, the build process works well.

.github/workflows/main.yml
```diff
@@ -100,9 +103,31 @@ jobs:
...
+
+    - run: |
+        curl -O https://www.nasm.us/pub/nasm/releasebuilds/${{ env.WINNASMVERSION }}/win64/nasm-${{ env.WINNASMVERSION }}-win64.zip
+        unzip nasm-${{ env.WINNASMVERSION }}-win64.zip
+        export PATH=$PWD/nasm-${{ env.WINNASMVERSION }}:$PATH
+        cargo test --manifest-path testcrate/Cargo.toml --target ${{ matrix.target }} -vv
+        cargo test --manifest-path testcrate/Cargo.toml --target ${{ matrix.target }} --release -vv
+        cargo run --release --target ${{ matrix.target }} --manifest-path testcrate/Cargo.toml --features package -vv
+      if: startsWith(matrix.os, 'windows') && startsWith(matrix.target, 'x86_64')
+      name: Run tests with nasm (Windows x86_64)
```
alexcrichton, line 129: Instead of being subsequent steps could these be two new matrix entries to allow them to run in parallel as well?

alexcrichton (commented):
Thanks! In addition to what's below could the env var being set to 1 assert that `nasm.exe` is used? That way if CI is misconfigured we'll catch it early.

KaneGreen:
Well, there so many tests about windows. 32bit/64bit, crt static yes/no, nasm none/installed/env-var -- 2*2*3 == 12 tests. Anyway,these tests passed.

alexcrichton:
Er well I don't think that the full matrix really needs to be tested. I think it's fine to just add a single x86_64 builder which enables nasm.

KaneGreen:
In this commit, I reduced the number of tests. Maybe only one left for nasm.exe may be too few.
In this version, we have 4 new tests for nasm.exe (comparing to e634b86cbfb6ef279b950a81ae65052d32d1468c)
|  | system | crt_static | nasm_exe | env_no_nasm | remark |
| -- | -- | -- | -- | -- | -- |
| 1 | i686 |  | installed |  | to test wheter it builds on win32 |
| 2 | x86_64 |  | installed |  | to test wheter it builds on win64 |
| 3 | x86_64 |  | installed | set | to test wheter the env var works |
| 4 | x86_64 | yes | installed |  | to test wheter it builds with crt_static |

KaneGreen:
This force push makes all previous commits as one commit.

alexcrichton:
Thanks for that but I'm not super interested in adding tons of CI configuration just for this use case, it's a lot of goop to maintain. I think that all that needs to be added is a singular new builder which has a singular extra step of downloading nasm. I then think that the env var, when set to `1` (or nonzero basically), would panic the build script of `nasm.exe` isn't detected. This new CI builder would enable the env var which guarantees that nasm is used and that it builds/works.

KaneGreen:
I'll reduce the numbers of CI tests.

However, I've changed the env var and now it won't panic. See https://github.com/alexcrichton/openssl-src-rs/pull/87#issuecomment-800758206 for details.

About the env var, the two methods:
- [OPENSSL_RUST_USE_NASM (comment)](https://github.com/alexcrichton/openssl-src-rs/pull/87#issuecomment-800356341)
- [OPENSSL_RUST_NO_NASM (comment)](https://github.com/alexcrichton/openssl-src-rs/pull/87#issuecomment-800758206)

Which would you like?

.github/workflows/main.yml
```diff
@@ -90,18 +102,37 @@ jobs:
...
+        unzip nasm-${WINNASMVERSION}-win64.zip
+        mv nasm-${WINNASMVERSION} nasm_win64
+    - name: Set NO_NASM env var (Windows)
+      if: startsWith(matrix.os, 'windows') && matrix.env_no_nasm == 'set'
+      run: echo OPENSSL_RUST_NO_NASM=1 >> $GITHUB_ENV
+    - name: Run tests (Windows)
+      if: startsWith(matrix.os, 'windows')
+      run: |
+        if [ '${{ matrix.nasm_exe }}' = 'installed' ] ; then
+        export PATH=$PWD/nasm_win64:$PATH
```
KaneGreen, line 125: Why do I add nasm to PATH in this step instead of in `Download nasm.exe`?
Because using `echo PATH=$PWD/xxxxxx:$PATH >> $GITHUB_ENV` in previous step will cause an error when configuring.

alexcrichton: For adding to `PATH` you'll want to use a different command -- https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-system-path

.github/workflows/main.yml
```diff
@@ -90,18 +102,37 @@ jobs:
...
+      if: "!startsWith(matrix.os, 'windows')"
+      run: |
         set -e
         cargo generate-lockfile
         ./ci/run-docker.sh ${{ matrix.target }}
-      if: "!startsWith(matrix.os, 'windows')"
-      name: Run tests (not Windows)
-    - run: |
+    - name: Download nasm.exe (Windows)
+      if: startsWith(matrix.os, 'windows') && matrix.nasm_exe == 'installed'
```
alexcrichton, line 112: I think you can remove the windows check here sce `nasm_exe` is only set for windows builders.

.github/workflows/main.yml
```diff
@@ -90,18 +102,37 @@ jobs:
...
         cargo test --manifest-path testcrate/Cargo.toml --target ${{ matrix.target }} --release -vv
         cargo run --release --target ${{ matrix.target }} --manifest-path testcrate/Cargo.toml --features package -vv
-      if: startsWith(matrix.os, 'windows')
-      name: Run tests (Windows)
+    - name: Verify no-nasm option (Windows)
+      if: startsWith(matrix.os, 'windows') && matrix.env_no_nasm == 'set'
+      run: |
+        # if the nasm.exe is disabled, there should be 'no-asm' in the configure command
+        echo 'CI: In output files, there should be no-asm configured:'
+        grep 'no-asm' target/${{ matrix.target }}/*/build/*/output
```
alexcrichton, line 135: I would prefer to avoid doing this since it's an unstable implementation detail of Cargo. As I've mentioned before I think the build script should panic if nasm is explicitly requested but it's not found.

KaneGreen: I see this output file path in the cargo book. So, I thought it is a stable feature. Now, I'll remove this.
However, on how to switch nasm.exe on and off, do you mean that we should use the [OPENSSL_RUST_USE_NASM](https://github.com/alexcrichton/openssl-src-rs/pull/87#issuecomment-800356341) way?

KaneGreen:
In this version, I set the environment variable back to `OPENSSL_RUST_USE_NASM`. But its behavior is changed:

function `is_nasm_ready()` return value:
| env var value | `yes` | `auto` or not set | `no` |
| -- | -- | -- | -- |
| nasm is installed | true ✔ | (auto-detect) true ✔ | false ❌ |
| nasm not installed | panic! 💥 | (auto-detect) false ❌ | false ❌ |

This information is also updated in `README.md`.

------------------------------------
About the CI tests:
Now, there is only one test abuot nasm: it only tests that the nasm works with this build process.
I didn't test the env var for the following reason.
1. I think the code for the `OPENSSL_RUST_USE_NASM` env var is simple enough.
2. I've tested these `is_nasm_ready()` and `check_env_var()` functions several times in my local Windows PC.
3. Testing the `OPENSSL_RUST_USE_NASM` env var needs more jobs in CI. And CI scripts will become more complicated.
------------------------------------
About the `panic` when nasm not installed and the env var is set to `yes`:
This panic is not caused by my code. It is caused by the `./Configure` not exit with code 0, which satisfies this requirement:
> As I've mentioned before I think the build script should panic if nasm is explicitly requested but it's not found.
[(source)](https://github.com/alexcrichton/openssl-src-rs/pull/87#discussion_r598767471)

src/lib.rs
```diff
@@ -65,6 +65,44 @@ impl Build {
...
     }
 
+    #[cfg(windows)]
+    fn check_env_var(&self, var_name: &str) -> Option<bool> {
+        env::var_os(var_name).and_then(|s| {
+            let tmp = s
+                .to_str()
+                .expect("The environment variable has invalid Unicode")
+                .to_lowercase();
+            match &tmp[..] {
```
alexcrichton, line 75: I don't think there's any need to go all out here, let's just check for either 0 or 1

.github/workflows/main.yml
```diff
@@ -90,18 +94,25 @@ jobs:
...
-      if: "!startsWith(matrix.os, 'windows')"
-      name: Run tests (not Windows)
-    - run: |
+    - name: Download nasm.exe (Windows)
+      if: matrix.nasm_exe == 'installed'
+      run: |
+        WINNASMVERSION='2.15.05'
+        curl -O https://www.nasm.us/pub/nasm/releasebuilds/${WINNASMVERSION}/win64/nasm-${WINNASMVERSION}-win64.zip
+        unzip nasm-${WINNASMVERSION}-win64.zip
+        echo "$PWD/nasm-${WINNASMVERSION}" >> $GITHUB_PATH
```
alexcrichton, line 109: Could this step also set `OPENSSL_RUST_USE_NASM`?

src/lib.rs
```diff
@@ -65,6 +65,44 @@ impl Build {
...
+                // On Windows, use cmd `where` command to check if nasm is installed
+                let wherenasm = Command::new("cmd")
+                    .args(&["/C", "where nasm"])
+                    .output()
+                    .expect("Failed to execute `cmd`.");
+                wherenasm.status.success()
+            })
+    }
+
+    #[cfg(not(windows))]
```
alexcrichton, line 100: I think the `#[cfg]` here can be avoided since nothing is windows-specific in that it only compiles on Windows.

src/lib.rs
KaneGreen: Well, there are there situation we need consider: force-enable, force-disable and auto-detecting. In this case, I think 0 or 1 is not suitable. I think 'yes/y', 'no/n', 'auto/a' will be better. (And leaving it unset means auto) (I'll remove those 'true'/'false', whatever)
 What do you think?

.github/workflows/main.yml
KaneGreen: Leaving it unset means auto-detecting. Adding this makes it works better in CI -- if some one don't install nasm correctly, the panic will let us know.

src/lib.rs
KaneGreen: I don't think these bothers. I think this notices us that we don't consider running nasm.exe on a non-windows os. (Even though nasm has non-windows version. However, according to OpenSSL's docs, it doesn't need nasm to build on non-windows platforms.)

src/lib.rs
alexcrichton: I personally don't really want to maintain an env var which is so meticulously parsed, I think a simple 0/1 is all we need. This isn't really going to be heavily used enough to warrant so much to handle it.

KaneGreen:
In this version, the acceptable values of the environment variable `OPENSSL_RUST_USE_NASM` is changed:

function `is_nasm_ready()` return value:
| env var value | `1` | not set | `0` |
| -- | -- | -- | -- |
| nasm is installed | true ✔ | (auto-detect) true ✔ | false ❌ |
| nasm not installed | panic! 💥 | (auto-detect) false ❌ | false ❌ |

This information is also updated in `README.md`.

KaneGreen:
Thank you for considering merging this PR. And thank you for your suggestions during this process.

Issue #30, opened 2019-04-11
go easy on the unwraps...


thedavidmeister:
especially in `cp_r` there are a lot of unwraps that can make debugging difficult

everywhere that works with the file system is easier to work with if it returns results

i ran into this when debugging file path and permissions issues on windows

PR #99, opened 2021-08-08
Handling unwrap errors in lib.rs

yaleman:
This blind unwrap seems to be a very common error in builds, when it panics it gives the end user no clear way of identifying what's wrong, this PR adds some handling to show the file path that can't be found.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -543,7 +543,7 @@ fn sanitize_sh(path: &Path) -> String {
             None => {
                 panic!("Failed to find path {:?}", &path);
             }
-        }
+        };
     }
 
     let path = match path.to_str() {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -538,7 +538,12 @@ fn apply_patches_musl(target: &str, inner: &Path) {
 
 fn sanitize_sh(path: &Path) -> String {
     if !cfg!(windows) {
-        return path.to_str().unwrap().to_string();
+        return match path.to_str() {
+            Some(value) => value.to_string(),
+            None => {
+                panic!("Failed to find path {:?}", &path);
+            }
+        }
     }
 
     let path = match path.to_str() {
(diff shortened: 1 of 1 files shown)


sfackler:
"Failed to find path" does not seem to be a very helpful message when the actual problem is a non-UTF8 path.

yaleman:
Turns out there was another one I was hitting, so I've updated that as well.

alexcrichton:
Are these unwraps getting hit frequently in practice? I think it's fine to get a better error message for when they are hit but if it's frequent then I'd want to fix the underlying issue ideally.

yaleman:
I had the issue, and looking in the issues for this project, it looks like the following tickets are related: #30, #47, #94. 

My issue was that I was missing `make`, which isn't mentioned in the error anywhere previously. The UTF path thing is much less of an issue than the error handling in `command.status().unwrap();` so I've updated the title of the PR.

src/lib.rs
```diff
@@ -474,7 +474,12 @@ impl Build {
 
     fn run_command(&self, mut command: Command, desc: &str) {
         println!("running {:?}", command);
-        let status = command.status().unwrap();
+        let status = match command.status() {
+            Ok(value) => value,
+            Err(error) => {
+                panic!("Error running command {:?} => {:?}", command, error);
```
alexcrichton, line 480: I don't think the command needs to be printed again because it was already printed a few lines above this?

src/lib.rs
```diff
@@ -538,9 +543,20 @@ fn apply_patches_musl(target: &str, inner: &Path) {
...
+        };
     }
-    let path = path.to_str().unwrap().replace("\\", "/");
+
+    let path = match path.to_str() {
+        Some(value) => value.replace("\\", "/"),
+        None => {
+            panic!("Failed to convert path to string: {:?}", &path);
+        }
+    };
```
alexcrichton, line 559: Could this and the above `to_str` use a helper function instead of duplicating?

yaleman: Probably

src/lib.rs
yaleman: Done

PR #102, opened 2021-09-09
Upgrade to OpenSSL v3.0.0

lupine:
On amd64, trying to compile a project (https://github.com/deltachat/deltachat-core-rust) that uses openssl-sys and vendors openssl via openssl-src, gets me this:

```
thread 'main' panicked at 'OpenSSL library directory does not exist: deltachat-core-rust/target/debug/build/openssl-sys-72260f2c868a930d/out/openssl-build/install/lib', openssl-sys/build/main.rs:66:9
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:515:5
     1: std::panicking::begin_panic_fmt
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:457:5
     2: build_script_main::main
     3: core::ops::function::FnOnce::call_once
  note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
This is solved by passing `--libdir=lib` to the configure script.

There's also a warning to do with `bin_dir`:

```
Compiling openssl-src v300.0.0+3.0.0 (/home/lupine/dev/github.com/deltachat/openssl-src-rs)
warning: field is never read: `bin_dir`
  --> /home/lupine/dev/github.com/deltachat/openssl-src-rs/src/lib.rs:25:5
   |
25 |     bin_dir: PathBuf,
   |     ^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```

But, it does compile and successfully vendor in openssl 3:

```
$ target/debug/examples/repl foo
OpenSSL 3.0.0 7 sep 2021
Delta Chat Core is awaiting your commands.
```

OpenSSL 3 has a better license for combination with GPL software that doesn't include the OpenSSL exemption, which is why I'm playing with this ^^

PR #103, opened 2021-09-09
Fix building against OpenSSL 3.0.0

lupine:
Unlike OpenSSL 1, OpenSSL 3 requires build.info files in its `test/` and `fuzz/` directories to be present. Without them, the configure script fails. They're excluded from the crate for space reasons; account for this in the build script by creating empty files to be included, since these features are disabled anyway.

More information here: https://github.com/alexcrichton/openssl-src-rs/pull/102

Signed-off-by: Nick Thomas <me@ur.gs>

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -8,6 +8,14 @@ if [ "$1" = "aarch64-apple-darwin" ] ; then
 	export CARGO_TARGET_AARCH64_APPLE_DARWIN_RUNNER=echo
 fi
 
+# Remove directories that are excluded by Cargo.toml
+rm -rf openssl/boringssl
+rm -rf openssl/fuzz
+rm -rf openssl/krb5
+rm -rf openssl/pyca-cryptography
+rm -rf openssl/test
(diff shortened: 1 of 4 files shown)


ci/run.sh
```diff
@@ -8,6 +8,14 @@ if [ "$1" = "aarch64-apple-darwin" ] ; then
 	export CARGO_TARGET_AARCH64_APPLE_DARWIN_RUNNER=echo
 fi
 
+# Remove directories that are excluded by Cargo.toml
```
lupine, line 11: I'm not too happy about duplicating these. Perhaps a better option would be to `cargo package`, then untar that and use it as the starting point for CI?

lupine: Hmm, the CI runs... read-only?

lupine: Ah, I see, we set the flag in run-docker.sh

I think I've got the `cargo package` approach working, I'll push it as a follow-up commit.

lupine:
Closed in favour of #104

PR #104, opened 2021-09-09
Fix building against OpenSSL 3.0.0 (take 2)

lupine:
Unlike OpenSSL 1, OpenSSL 3 requires build.info files in its fuzz/ directory to be present. Without it, the configure script fails. `fuzz/` is excluded from the crate for space reasons; this change puts it back and excludes `doc/` and `demos/` instead, to get back below the 10MiB limit.

More information here: #102

This PR is like https://github.com/alexcrichton/openssl-src-rs/pull/103 but tries to use `cargo package` to make CI closer to a real build, rather than just `rm`ing a bunch of files in the main directory.

Signed-off-by: Nick Thomas me@ur.gs

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -8,16 +8,17 @@ if [ "$1" = "aarch64-apple-darwin" ] ; then
 	export CARGO_TARGET_AARCH64_APPLE_DARWIN_RUNNER=echo
 fi
 
-# Remove directories that are excluded by Cargo.toml
-rm -rf openssl/boringssl
-rm -rf openssl/fuzz
-rm -rf openssl/krb5
-rm -rf openssl/pyca-cryptography
-rm -rf openssl/test
-rm -rf openssl/wycheproof
+# Use cargo package to ensure we don't rely on any files excluded by Cargo.toml
+cargo package --allow-dirty
 
-cargo test --manifest-path testcrate/Cargo.toml --target $1 -vv
-cargo test --manifest-path testcrate/Cargo.toml --target $1 -vv --release
+version=$(cargo run output-version)
+
+testcrate_dir="$(pwd)/testcrate"
+
+cd "target/package/openssl-src-$version"
+
+cargo test --manifest-path "$testcrate_dir/Cargo.toml" --target $1 -vv
+cargo test --manifest-path "$testcrate_dir/Cargo.toml" --target $1 -vv --release
 if [ "$1" = "x86_64-unknown-linux-gnu" ] ; then
-	cargo test --manifest-path testcrate/Cargo.toml --target $1 -vv --all-features
+	cargo test --manifest-path "$testcrate_dir/Cargo.toml" --target $1 -vv --all-features
 fi
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -1,4 +1,5 @@
 target=$1
+testcrate_dir="$(pwd)/testcrate"
 set -ex
 
 if [ "$1" = "aarch64-apple-darwin" ] ; then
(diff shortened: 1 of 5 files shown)


lupine:
OK @alexcrichton, I've got a choice between two PRs for you.

Both use exactly the same fix for the actual build - creating `fuzz/build.info` and `test/build.info` files before running `./Configure`, which is sufficient to get it to compile without those directories (so the `exclude` directives can stay as they are).

Where they differ is in how they modify CI - #103 duplicates the list of excluded files, makes the code directory RW, and `rm -rf`s those directories prior to running the tests.

This PR (which I prefer) uses `cargo package` to mirror the contents to `target/package/openssl-src-<version>`  before running the tests. The packaged version lacks the `testcrate` directory as well as the excluded paths.

It does mean we need to know the version of openssl-src that we're building, to construct the path. This is in Cargo.toml, but how to access it in a way that's sensible across all platforms is a bit of a vexed question :sweat_smile:

It's in `cargo metadata`, but that requires `jq` to parse it out, which isn't installed in CI. I didn't fancy trying to work out how to get access to it in windows.

Another option would be to grep | cut | tr the Cargo.toml file. I didn't try this one.

When we `cargo run`, it's provided as an envvar to the binary that gets built and executed. I took this approach, adding `src/bin/output_version.rs`. It's a bit weird, but it works.

Or we can just merge #103 :sweat_smile:

sfackler:
This approach seems cleaner to me as well.

alexcrichton:
I agree that this is better than hardcoding lists, but build scripts ideally shouldn't be modifying the contents of the directories that they're unpackaged into. I'm also wary to delete parts of OpenSSL's source tree that it actually needs to build and trying to fake our way to adding them back in.

Are there perhaps more targeted exclusions of files which don't affect the build system but still allow us to get the tarball under the 10MB limit?

sfackler:
We could also ask Crates.io to bump the limit for us. It seems like the official tarball off of openssl.org is ~14mb as well.

lupine:
@alexcrichton hmm, I note that until just recently we were modifying the source code directly, not just the build system ;) 

However, I've worked out that we can actually make `test/` unnecessary, by using `no-tests` (instead of the former `no-unit-tests`). There's no way exclude `fuzz/` that I can see, but leaving out `test/` gets us back down to 11MiB.

I've just pushed a commit that takes out `demos/` and `doc/` as well, let's see if CI passes with it.

lupine:
If we're all agreed that the CI changes are best like this, I@ll close #103

lupine:
OK, this looks like it would do the trick @alexcrichton . WDYT?

```
openssl-src-rs$ du -bs target/package/*
60262607	target/package/openssl-src-300.0.0+3.0.0
9800148	target/package/openssl-src-300.0.0+3.0.0.crate
```

ci/run.sh
```diff
@@ -8,8 +8,17 @@ if [ "$1" = "aarch64-apple-darwin" ] ; then
...
-cargo test --manifest-path testcrate/Cargo.toml --target $1 -vv
-cargo test --manifest-path testcrate/Cargo.toml --target $1 -vv --release
+# Use cargo package to ensure we don't rely on any files excluded by Cargo.toml
+cargo package --allow-dirty
+
+version=$(cargo run output-version)
+
+testcrate_dir="$(pwd)/testcrate"
+
+cd "target/package/openssl-src-$version"
```
alexcrichton, line 18: To avoid adding a new binary just to print the version, could this do `cd target/package/openssl-src-*`?

ci/run.sh
```diff
@@ -8,8 +8,17 @@ if [ "$1" = "aarch64-apple-darwin" ] ; then
 	export CARGO_TARGET_AARCH64_APPLE_DARWIN_RUNNER=echo
 fi
 
-cargo test --manifest-path testcrate/Cargo.toml --target $1 -vv
-cargo test --manifest-path testcrate/Cargo.toml --target $1 -vv --release
+# Use cargo package to ensure we don't rely on any files excluded by Cargo.toml
+cargo package --allow-dirty
```
alexcrichton, line 12: Could this be split out to a separate CI step for one platform to test the packaged version? I'd prefer to still do at least some testing of the in-tree version on CI.

alexcrichton (commented):
Seems reasonable to me!

ci/run.sh
lupine: Hmm. Even in the naïve case where we're starting with a clean `target/` directory we have two matching entries (the directory and the .crate archive), so this will fail. We can delete *.crate to make it work, however. We should ensure `target/package` is emptied first though, since it relies on the assumption that we're starting with an empty directory.

ci/run.sh
lupine: @alexcrichton I had a look, but I'm not familiar enough with GitHub workflows to make this happen :grimacing: .

As a halfway house, I've moved the logic to be within the `if [ "$1" = "x86_64-unknown-linux-gnu" ]` block, which has the same effect but without it being a separate CI step. What do you think?

ci/run.sh
lupine: I like `target/ci` so I've just done that

alexcrichton:
Thanks, looks good to me!

PR #106, opened 2021-09-11
Add support for UWP targets

bdbai:
Since Universal Windows Platform (UWP) App is supported in OpenSSL 3.0.0 ([Build Configuration](https://github.com/openssl/openssl/blob/master/Configurations/50-win-onecore.conf)), we can make use of it by mapping UWP targets to corresponding configuration entries.

BTW, cross compilation in Windows failed because of too long paths exceeding 260 characters. Is there any workaround?


ValHeimer:
> BTW, cross compilation in Windows failed because of too long paths exceeding 260 characters. Is there any workaround?

Ok I'm not the only one having this error...
I temporary fixed it by setting the CARGO_TARGET_DIR env var to a shorter folder than my repo's one.
```bash
CARGO_TARGET_DIR=/c/target cargo build
```

alexcrichton:
Thanks! And yeah for the long path issue @ValHeimer's suggestion is probably the easiest to work around it.

PR #107, opened 2021-09-13
Try excluding the fuzz corpora from crates.io

alexcrichton:
Otherwise we're still sitting at 12M which is too big for crates.io


sfackler:
Might be worth adding a CI step to run `cargo package` and check the tarball size.

PR #112, opened 2021-11-27
Support riscv64gc-unknown-freebsd

t6:
Also see https://github.com/rust-lang/rust/pull/91284


alexcrichton:
Seems reasonable to me! If you wouldn't mind pinging this when the upstream rust-lang/rust PR has merged I'll merge this here as well.

t6:
It has been merged. Thanks!

PR #113, opened 2021-12-14
Support compiling to wasm32-wasi target

mdrokz:
Adds support for compiling to wasm32-wasi target

# Changes

- added the usage of `wasiconfigure` and `wasimake` when targeting wasm
- added wasm-wasi specific config options

# Issues

- https://github.com/sfackler/rust-openssl/issues/1016

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -218,8 +218,9 @@ impl Build {
         }
 
         if target.contains("wasm") {
+            // -no-threads we dont have thread support in WASI
             configure.arg("no-threads");
-            configure.arg("no-asm");
+            // -no-sock we dont have sockets in WASI
             configure.arg("no-sock");
             configure.arg("no-afalgeng");
             configure.arg("-DOPENSSL_SYS_NETWARE");
@@ -227,6 +228,7 @@ impl Build {
             configure.arg("-DSIG_IGN=0");
             configure.arg("-DHAVE_FORK=0");
             configure.arg("-DOPENSSL_NO_AFALGENG=1");
+            // --with-rand-seed=getrandom (needed to force using getentropy because WASI has no /dev/random or getrandom)
             configure.arg("--with-rand-seed=getrandom");
         }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,6 +53,7 @@ impl Build {
 
     fn cmd_make(&self) -> Command {
         let host = &self.host.as_ref().expect("HOST dir not set")[..];
+        let target = &self.target.as_ref().expect("TARGET dir not set")[..];
         if host.contains("dragonfly")
             || host.contains("freebsd")
             || host.contains("openbsd")
@@ -61,7 +62,13 @@ impl Build {
         {
             Command::new("gmake")
         } else {
-            Command::new("make")
+            if target.contains("wasm") {
+                let mut cmd = Command::new("wasimake");
+
+                std::mem::replace(cmd.arg("make"),Command::new("wasimake"))
+            } else {
+                Command::new("make")
+            }
         }
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -217,6 +217,19 @@ impl Build {
             configure.arg("no-stdio");
         }
 
+        if target.contains("wasm") {
+            configure.arg("no-threads");
+            configure.arg("no-asm");
+            configure.arg("no-sock");
+            configure.arg("no-afalgeng");
+            configure.arg("-DOPENSSL_SYS_NETWARE");
+            configure.arg("-DSIG_DFL=0");
+            configure.arg("-DSIG_IGN=0");
+            configure.arg("-DHAVE_FORK=0");
+            configure.arg("-DOPENSSL_NO_AFALGENG=1");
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -61,7 +62,13 @@ impl Build {
         {
             Command::new("gmake")
         } else {
-            Command::new("make")
+            if target.contains("wasm") {
```
mzrokz, line 65: You can wrap this in a single else if statement

src/lib.rs
```diff
@@ -61,7 +62,13 @@ impl Build {
...
+                let mut cmd = Command::new("wasimake");
+
+                std::mem::replace(cmd.arg("make"), Command::new("wasimake"))
```
alexcrichton, line 68: What is `wasimake` and `wasiconfigure`? I don't think these are provided by OpenSSL and otherwise I don't want to rely on too many external programs. (I've also never heard of these myself)

src/lib.rs
```diff
@@ -155,7 +169,7 @@ impl Build {
             // Avoid multilib-postfix for build targets that specify it
             .arg("--libdir=lib");
 
-        if cfg!(not(feature = "legacy")) {
+        if cfg!(not(feature = "legacy")) && !target.contains("wasm") {
```
alexcrichton, line 172: Why is this condition here added for wasm?

src/lib.rs
```diff
@@ -203,6 +217,19 @@ impl Build {
...
+            configure.arg("no-threads");
+            configure.arg("no-asm");
+            configure.arg("no-sock");
+            configure.arg("no-afalgeng");
+            configure.arg("-DOPENSSL_SYS_NETWARE");
+            configure.arg("-DSIG_DFL=0");
+            configure.arg("-DSIG_IGN=0");
+            configure.arg("-DHAVE_FORK=0");
+            configure.arg("-DOPENSSL_NO_AFALGENG=1");
+            configure.arg("--with-rand-seed=getrandom");
```
alexcrichton, line 230: Can you add some comments for what's going on here for wasm?

alexcrichton (commented):
Thanks! Can you add a builder to CI so we can track whether it continues to work over time or not?

src/lib.rs
mdrokz: > What is `wasimake` and `wasiconfigure`? I don't think these are provided by OpenSSL and otherwise I don't want to rely on too many external programs. (I've also never heard of these myself)

Yes these commands are included in the `wasienv` toolchain https://docs.wasmer.io/ecosystem/wasienv and right now i only found this way to compile openssl for wasm.

src/lib.rs
mdrokz: > Why is this condition here added for wasm?

Ah my bad, i accidentally left that condition there.

src/lib.rs
mdrokz: > Can you add some comments for what's going on here for wasm?

Sure

mdrokz:
> Thanks! Can you add a builder to CI so we can track whether it continues to work over time or not?

OpenSSL 3 isnt compatible with wasm32-wasi target by default, in order to compile it you need to apply patches and i only found patches for OpenSSL 1.1.1d and compiled it, but the openssl-src-rs repo always fetches the latest openssl, is there any way to fetch only OpenSSL 1 for the builder?

src/lib.rs
alexcrichton: Can this patch be written without requiring those tools? Without knowing what they do it seems like they could introduce hidden dependencies we aren't aware of.

src/lib.rs
mdrokz: > No this repository only has one version of OpenSSL via a submodule, which is currently 3+

Yes this repo only contains openssl 3, but right now only OpenSSL 1.1.1d is supported with specific patches for wasm32-wasi, see https://github.com/alexcrichton/openssl-src-rs/pull/113#issuecomment-993719338

src/lib.rs
mdrokz: > Can this patch be written without requiring those tools? Without knowing what they do it seems like they could introduce hidden dependencies we aren't aware of.

I dont think so, but this only requires the user to install the wasienv toolchain.

src/lib.rs
alexcrichton: If that's the case then I don't think this can land right now.

src/lib.rs
alexcrichton: Personally I do not want to land this with a dependency on the wasienv toolchain. I don't know why it's necessary as opposed to the standard wasi-sdk.

mdrokz: > Personally I do not want to land this with a dependency on the wasienv toolchain. I don't know why it's necessary as opposed to the standard wasi-sdk.

I tried compiling using the standard wasi-sdk but it didnt work no matter what i tried, this seems to be the proper way to configure the build for wasi as far as i know.

src/lib.rs
mdrokz: > If that's the case then I don't think this can land right now.

I will try patching the latest OpenSSL 3 and see if this works with that or not.

mdrokz:
Closed because of #119

PR #116, opened 2021-12-14
Fix aarch64-apple-darwin CI

ehuss:
These lines should no longer be necessary.  The default XCode is now 13.1 (and I believe the default SDK is macos 12.0).  I believe these were added since at the time the default xcode did not support aarch64-apple-darwin.


alexcrichton:
Thanks!

PR #119, opened 2022-01-24
Added support for wasm32-wasi target

OtaK:
Hi! 

I'm currently on my way to make SQLite work with `wasm32-wasi`.
So, one of the blockers was OpenSSL in general since SQLite depends heavily on it.

Here's my shot at it: 

* Introduces no new dependencies
* Small alteration of the docker test harness to: 
	* install `wasmtime` & set it up as test runner
	* install & setup wasi-sdk
* Compatible with all versions of OpenSSL (1.1.x and 3.x.x were tested).

This should be the first step in closing https://github.com/sfackler/rust-openssl/issues/1016

Should also supersede #113

Please give me any feedback! I'll tend to it ASAP.


testcrate/src/lib.rs
```diff
@@ -6,10 +6,15 @@ extern "C" {
...
+mod tests {
+    use super::OpenSSL_version_num;
+
+    #[test]
+    fn version_works() {
+        unsafe {
+            println!("{:#x}", OpenSSL_version_num());
+            assert!(OpenSSL_version_num() > 0);
+        }
     }
```
alexcrichton, line 19: I think this file change may not be necesary?

.github/workflows/main.yml
```diff
@@ -59,9 +59,10 @@ jobs:
         - target: x86_64-pc-windows-gnu
           rust: stable
           os: ubuntu-latest
-        # - target: asmjs-unknown-emscripten
-        #   rust: stable
-        #   os: ubuntu-latest
+        - target: wasm32-wasi
+          rust: stable
+          os: ubuntu-latest
+          crt_static: yes
```
alexcrichton, line 65: Is this `crt_static` config necessary?

ci/run.sh
```diff
@@ -8,8 +8,15 @@ if [ "$1" = "aarch64-apple-darwin" ] ; then
   export CARGO_TARGET_AARCH64_APPLE_DARWIN_RUNNER=echo
 fi
 
-cargo test --manifest-path "$testcrate_dir/Cargo.toml" --target $1 -vv
-cargo test --manifest-path "$testcrate_dir/Cargo.toml" --target $1 -vv --release
+if [ "$1" = "wasm32-wasi" ] ; then
```
alexcrichton, line 11: Could this move from the `run.sh` script to a dockerfile where the platform-specific runners and such live?

src/lib.rs
```diff
@@ -423,6 +423,56 @@ impl Build {
...
+                    // Termios isn't available whatsoever on WASM/WASI so we disable that
+                    // fatal error: 'termios.h' file not found
+                    "no-ui-console",
+                    // WASI doesn't support UNIX sockets so we preemptively disable it
+                    "no-sock",
+                    // WASI doesn't have a concept of syslog, so we disable it
+                    // fatal error: 'syslog.h' file not found
+                    "-DNO_SYSLOG",
+                    // wasm lacks signal support; to enable minimal signal emulation, compile with
+                    // -D_WASI_EMULATED_SIGNAL and link with -lwasi-emulated-signal
```
alexcrichton, line 437: If these `-l` flags are required with these compilation flags I think that the build script here and/or the build results will need to print out appropriate cargo directives to indicate that these extra libraries should be linked.

src/lib.rs
```diff
@@ -423,6 +423,56 @@ impl Build {
...
+                    "no-afalgeng",
+                    "-DOPENSSL_NO_AFALGENG=1",
+                    // ? Do we really want to always build statically for wasm? I guess yes?
+                    "static",
+                    // No fork(1) support in WASI, so we disable it
+                    "-DHAVE_FORK=0",
+                    "-DOPENSSL_NO_SPEED=1",
+                    "-DOPENSSL_SYS_NETWARE",
+                    "-DSIG_DFL=0",
+                    "-DSIG_IGN=0",
```
alexcrichton, line 472: Are all of these flags required? Can some be removed and the build still work?

src/lib.rs
```diff
@@ -423,6 +423,56 @@ impl Build {
...
+                    // WASI lacks process identifiers; to enable emulation of the `getpid` function using a
+                    // placeholder value, which doesn't reflect the host PID of the program, compile with
+                    // -D_WASI_EMULATED_GETPID and link with -lwasi-emulated-getpid
+                    "-D_WASI_EMULATED_GETPID",
+                    // ----------
+                    // The following arguments are taken from:
+                    // https://github.com/wapm-packages/OpenSSL/blob/master/build.sh#L21
+                    // Since this is a successful and reliable source for configuring WASM/I builds.
+                    // I will try to explain those to the best of my ability though.
+                    // ----------
```
alexcrichton, line 455: Personally I think it's best to only list here what we need and explain the reason for adding them, so I think it's ok to remove this reference since each flag below should have comments as to why it's there.

src/lib.rs
```diff
@@ -423,6 +423,56 @@ impl Build {
...
+                    // WASI doesn't support (p)threads. Disabling preemptively.
+                    "no-threads",
+                    // WASI/WASM aren't really friends with ASM, so we disable it as well.
+                    "no-asm",
+                    // Disables the AFALG engine (AFALG-ENGine)
+                    // Since AFALG depends on `AF_ALG` support on the linux kernel side
+                    // it makes sense that we can't use it.
+                    "no-afalgeng",
+                    "-DOPENSSL_NO_AFALGENG=1",
+                    // ? Do we really want to always build statically for wasm? I guess yes?
```
alexcrichton, line 465: Yes, for wasm everything is always static right now. I don't know what effect this has on OpenSSL's build system though.

testcrate/src/lib.rs
OtaK: I can revert it if needed -- I just thought it was weird to have a test not wrapped in the usual conditional compilation directive/mod combo. Kinda instinctive fix 🤔

.github/workflows/main.yml
OtaK: From my experiments with WASI yes it is pretty much required all the time; I can try to disable it and see if things still work though.

ci/run.sh
OtaK: Yep! Seeing the WASI CI test fail, I started doing exactly that about the *test dependencies*. 
Adding to that, should I just pull out the `wasm32-wasi` target from the run.sh matrix instead and as you say, make it fully self-contained in a `Dockerfile`?

src/lib.rs
OtaK: I'll add those to be emitted for the sake of completeness. But from my experience playing with WASI, usually only the compilation directives are really required. (I get the "unused linker-arg" warning saying that those flags are ignored altogether most of the time)

src/lib.rs
OtaK: They can be removed yes, and the build would still build, but *working as intended* is something I'm not sure about.

I might try to dig into the wapm build to see if there's a rationale behind those flags.

alexcrichton:
Ah ok sorry I didn't realize that this wasn't passing CI, I'll take another look when CI is passing for wasm32-wasi

src/lib.rs
OtaK: Update: I tried to remove them. Build seems to work fine!

src/lib.rs
OtaK: Update: Those flags are now emitted but it seems this feature requires `-Zextra-link-arg` on `cargo`. 

I get the following warnings now: 
```
warning: cargo:rustc-link-arg requires -Zextra-link-arg flag
```

And again, the build & test run works fine without those link args. Only the compilation directives are a hard requirement.

src/lib.rs
OtaK: Done!

src/lib.rs
OtaK: Turns out providing `static` in this location as a cfg arg to OpenSSL prevents compilation as it would be provided *twice*. I removed it.

.github/workflows/main.yml
OtaK: Update: It was not needed. Pretty much the opposite, my bad!

testcrate/src/lib.rs
OtaK: Reverted!

src/lib.rs
dheater: Seems to me that it would be cleaner to omit the unnecessary linker flags and avoid the warnings.

OtaK: I'd favor this approach too; Also conceptually linker flags are usually the end consumer (= binary-producing)'s responsibility, so I don't think we need to bother with linkage in general.

ci/docker/wasm32-wasi/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+  gcc \
+  xz-utils \
+  libc6-dev
+
+RUN bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)" -- 14
+RUN update-alternatives --install /usr/bin/llvm-config llvm-config /usr/bin/llvm-config-14 14
+RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-14 14
+
+RUN curl https://wasmtime.dev/install.sh -sSf | bash
+RUN cargo install cargo-wasi
```
alexcrichton, line 18: I don't think that `cargo wasi` is really buying anything over what's already present in this CI, so can this be removed?

src/lib.rs
alexcrichton: If a warning is printed out about the library not being needed then presumably it's not needed? In that case it should be fine to just remove the comment here about the need for the `-l` on the library.

ci/docker/wasm32-wasi/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+  software-properties-common lsb-release gpg-agent gnupg gnupg1 gnupg2 \
+  make \
+  perl \
+  gcc \
+  xz-utils \
+  libc6-dev
+
+RUN bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)" -- 14
+RUN update-alternatives --install /usr/bin/llvm-config llvm-config /usr/bin/llvm-config-14 14
+RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-14 14
```
alexcrichton, line 15: Is this needed since wasi-sdk has its own version of clang?

ci/docker/wasm32-wasi/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM rust:1-slim
```
alexcrichton, line 1: I think it's ok to use the same base as the other images if the installation of `cargo-wasi` is removed.

ci/docker/wasm32-wasi/Dockerfile
OtaK: That's probably an old try of someting else altogether. I'll remove it.

ci/docker/wasm32-wasi/Dockerfile
OtaK: Cargo wasi is taking care of setting up the sysroot, test runner, target args and probably a plenty of other things. I wouldn't want to mess with that personally. It would amount to reimplementing cargo-wasi ourselves, I don't see the point IMO.

All of this is necessary so far

ci/docker/wasm32-wasi/Dockerfile
OtaK: Done by moving the cargo-wasi install to the github workflow

It was not really working anyway since the `docker run` imports cargo & rustc from the parent container to the child and I was setting up the inner container. Probably a side-effect of container caching that was making sure things work. It probably wouldn't have worked on a cold build.

src/lib.rs
OtaK: I did it with `println!("cargo:rustc-link-lib=flag-without-dash-L")`; Not sure what it does in practice but no warnings are printed neither from clang, rustc or lld whatsoever. I guess this is the correct way of doing things?

.github/workflows/main.yml
```diff
@@ -88,6 +88,9 @@ jobs:
     - name: Install Rust (rustup)
       run: rustup update ${{ matrix.rust }} --no-self-update && rustup default ${{ matrix.rust }}
     - run: rustup target add ${{ matrix.target }}
+    - name: Install Cargo-Wasi
+      if: matrix.target == 'wasm32-wasi'
+      run: cargo install cargo-wasi
```
alexcrichton, line 93: Sorry but what I meant was can this be removed entirely? I don't think `cargo wasi` is needed here

src/lib.rs
alexcrichton: I believe what's implemented in this PR right now, which is to do `println!("cargo:rustc-link-lib=wasi-emulated-signal");`, is what needs to be done. Does that print warnings or is there a raeson to not do that?

If the prints below stay then the comments here about libraries to link can probably be removed or just indicate that the corresponding library is linked below.

OtaK: No warnings, and indeed that's what I'm doing. I'll update the comments to reflect that.

.github/workflows/main.yml
OtaK: I removed cargo-wasi as per your request, seems to be building & running fine. But we lost the auto-`wasm-opt` in release mode. Seems fine to me for a CI.

alexcrichton:
Looks good to me, thanks!

PR #120, opened 2022-01-27
Fix weak-crypto feature

eaufavor:
These flags are disabled by default at least in newer versions of openssl.
So They have to be explicitly enabled to be used.

I've verified this behavior on openssl 1.1.1h and 3.0.1. Other flags/features might have the same issue, but I have no check yet.


alexcrichton:
Seems reasonable to me, thanks! While you're at it would you be ok inverting the conditional and swapping the cases? (seems a bit easier to read to me at least)

eaufavor:
Sure thing. Done.

Issue #127, opened 2022-03-16
Unable to generate static binaries on Haiku


kallisti5:
Haiku R1/beta3, Rust 1.57.0

Attempting to build a static openssl fails:
```
RUSTFLAGS='-C target-feature=+crt-static' cargo build --release
.
.
error: failed to run custom build command for `openssl-sys v0.9.72`
.
thread 'main' panicked at 'don't know how to configure OpenSSL for x86_64-unknown-haiku', /boot/home/.cargo/registry/src/github.com-1ecc6292.../openssl-src-111.17.0+1.1.1m/src/lib/rs:283:18
```

```
pkg-config openssl --static --libs --cflags
-I/packages/openssl-1.1.1m-1/.self/develop/headers -L/packages/openssl-1.1.1m-1/.self/develop/lib -lssl -lcrypto -lnetwork
```

kallisti5:
resolved via PR

PR #128, opened 2022-03-16
configs: Add current Haiku openssl configs

kallisti5:


kallisti5:
related: https://github.com/alexcrichton/openssl-src-rs/issues/127

PR #130, opened 2022-03-16
[111] Switch riscv64gc-unknown-freebsd to BSD-riscv64

pkubaj:


alexcrichton:
Thanks for the PR but I'd prefer to keep this branch as security fixes only. Is there a pressing reason to merge this though?

pkubaj:
There's no pressing reason, it's just that support for BSD-riscv64 was added in 1.1.1n, so I also wanted to add it here.

If you're against it, then ok.

PR #135, opened 2022-04-19
Add aarch64-apple-ios-sim target to 111 release

simlay:
I'm not sure this is the preferred flow for adding this change to the 111 release but figured I could create an issue to ask or just do it.


simlay:
After some looking around, it looks like the `release/111` branch is for security fixes as mentioned in https://github.com/alexcrichton/openssl-src-rs/pull/130#issuecomment-1069625131.

PR #136, opened 2022-05-05
Bump to 1.1.1o

amousset:


amousset:
I checked it with current version to be sure and it fails too, so I'm disabling it.

amousset:
I'll have a look at what is included

amousset:
It happens because of the new `wycheproof` submodule (which seems to be initialized automatically when updating the openssl submodule), when cloning `openssl-src-rs` from scratch and non-recursively initializing submodules, I get a size similar to previous releases.

alexcrichton:
Ah ok that was [fixed on `main`](https://github.com/alexcrichton/openssl-src-rs/pull/110) awhile back. Would you be interested in backporting that?

amousset:
Sure!

PR #137, opened 2022-05-05
Bump to 3.0.3

amousset:


alexcrichton:
Looks like the android build here is broken as well, and like the 1.1.1 update this'll either need to be fixed or removed here to land.

PR #139, opened 2022-05-17
Backport wycheproof exclude to 1.1.1 branch

amousset:


alexcrichton:
Thanks for the PR, but for the continued maintenance of this branch I think it would be good to backport the CI bits as well to protect against cases like this in the future. Would you be able to do that here as well?

alexcrichton:
Thanks for the PR, but for the continued maintenance of this branch I think it would be good to backport the CI bits as well to protect against cases like this in the future. Would you be able to do that here as well?

alexcrichton:
Thanks!

PR #150, opened 2022-07-20
Enable engine support for musl

neuronull:
In one of our (vector.dev) project's dependencies (rdkafka), a hard dependency on openssl Engine support was added.
This resulted in cross compilation errors for our project (since openssl-src was not enabling the engine support), the findings were documented here:

https://github.com/vectordotdev/vector/pull/10302

In particular, I found that if I made the below change to openssl-src , I was able to successfully compile to the target `x86_64-unknown-linux-musl` using `cross`.

Would this be an acceptable change to make in openssl-src with this knowledge ?

Thanks!

### DIFF
--- a/ci/docker/x86_64-unknown-linux-musl/Dockerfile
+++ b/ci/docker/x86_64-unknown-linux-musl/Dockerfile
@@ -6,4 +6,10 @@ RUN apt-get update -y && apt-get install -y --no-install-recommends \
   perl \
   gcc \
   libc6-dev \
-  musl-tools
+  musl-tools \
+  linux-headers-4.15.0-20-generic
+
+# link up the musl headers
+RUN ln -s /usr/include/x86_64-linux-gnu/asm /usr/include/x86_64-linux-musl/asm \
+  && ln -s /usr/include/asm-generic /usr/include/x86_64-linux-musl/asm-generic \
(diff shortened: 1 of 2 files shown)


neuronull:
Hmm , looks like the CI for musl failed :/ 
It appears to be installing the debian packages for musl directly.

alexcrichton:
Ah yeah, that's the reason it's disabled right now. If it can be enabled that would work well but it may require fiddling with cross-compile flags or something like that (I don't know myself)

neuronull:
I'll take a look and see if I can find a solution.

alexcrichton:
Unfortunately I don't think the right fix is to fix the container the build happens in here. Otherwise anyone else is going to have to apply the same fix which is a breaking change relative to today.

If you'd like you can also add a boolena parameter to configuration to opt-in to the `engine` being built.

neuronull:
> Unfortunately I don't think the right fix is to fix the container the build happens in here. Otherwise anyone else is going to have to apply the same fix which is a breaking change relative to today.
> 
> If you'd like you can also add a boolena parameter to configuration to opt-in to the `engine` being built.

Oh yes... that makes sense. I'll revert the CI change and add an configuration param 👍

neuronull:
Realized I need to do this for the 111 release

PR #151, opened 2022-07-21
add force-engine feature for enabling engine module

neuronull:
In one of our (vector.dev) project's dependencies (rdkafka), a hard dependency on openssl Engine support was added.
This resulted in cross compilation errors for our project (since openssl-src was not enabling the engine support), the findings were documented here:

https://github.com/vectordotdev/vector/pull/10302

Allowing this engine feature support to be configurable would help unblock us.

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -185,10 +185,13 @@ impl Build {
         }
 
         if target.contains("musl") || target.contains("windows") {
-            // This actually fails to compile on musl (it needs linux/version.h
+            // Engine module fails to compile on musl (it needs linux/version.h
             // right now) but we don't actually need this most of the time.
             // API of engine.c ld fail in Windows.
-            configure.arg("no-engine");
+            // Disable engine module unless force-engine feature specified
+            if !cfg!(feature = "force-engine") {
+                configure.arg("no-engine");
+            }
         }
 
         if target.contains("musl") {


alexcrichton:
I'd recommend if you can doing an end-to-end integration test with this. This PR, for example, doesn't add a Cargo feature so there's no easy way to actually turn on this `cfg` other than with `RUSTFLAGS` which is probably not what you intended.

neuronull:
Ah yes... I think that is all I was missing.
I just completed cross compiling our project to musl, and patching the `openssl-sys` crate to use this fork and the new feature, and it succeeded.

neuronull:
Thank you! 🙏

havasd:
Could you please elaborate how does this change supposed to work? I have a similar use-case as well.
But based on the way it looks it is unlikely that this is a suitable solution. At least I couldn't get it to work.

Because based on the usages this crate is always built in other crates. I.e.: In case of openssl-sys they are calling directly the Build in lib.rs. So I don't think it goes through the normal means of cargo build facilities (build.rs). I think this feature-flag would only be controllable in `Cargo.toml` if we would have the normal build procedure. (correct me if I am wrong)

Ideally I could think 2 ways to fix this problem:
- Export this feature all the way and make it possible to customize the Build section from openssl-sys
- Add sample test programs which can decide wether the current platform supports such facilities which are disabled by default on `musl` or `windows` (probably this one is more robust)

What do you think?

PR #162, opened 2022-10-14
Fetch openssldir from env variable

### LINKED ISSUE #161: Suggestion to fetch openssldir from env variable
I'm working on building a package that depends on `openssl-sys` using nix, which does not place openssl in `/usr/local/ssl`. Could this line check to see if `$OPENSSL_DIR` is set before it uses the above as default?

https://github.com/alexcrichton/openssl-src-rs/blob/048d5f5584643b02ed633ca0b2ac061188ad6f87/src/lib.rs#L151

Thank you

h7x4:
Resolves #161

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,7 +148,8 @@ impl Build {
         if target.contains("windows") {
             configure.arg("--openssldir=SYS$MANAGER:[OPENSSL]");
         } else {
-            configure.arg("--openssldir=/usr/local/ssl");
+            let openssldir = env::var("OPENSSL_DIR").unwrap_or("/usr/local/ssl".to_string());
+            configure.arg(&format!("--openssldir={}", openssldir));
         }
 
         configure


alexcrichton:
Instead of making this an env var can this be added as a configuration option on the builder?

DanielSidhion:
As far as I understand, adding this as a configuration option on the builder would imply that every crate that uses this one would need to properly configure it, is that correct? This will still require extra effort to fix every crate that depends on this, because when building with nix (as per #161) there will never be anything in `/usr/local/ssl`.

I don't know how many crates depend on this, but I think using the env var aligns with what the [`openssl`](https://docs.rs/openssl/latest/openssl/#manual) crate does, which would fix the issue.

I'm struggling with this when trying to package a crate with vendored openssl with nix, and would like to help get this merged.

alexcrichton:
AFAIK it's primarily `openssl-sys`, and maybe only `openssl-sys`, that depends on this crate. If the `openssl` crate has a convention for this environment variable then that's a good place to configure a builder-based option.

DanielSidhion:
It looks like there are a [few other dependents](https://crates.io/crates/openssl-src/reverse_dependencies), but the number is low enough that I agree a builder-based option is best. @h7x4 would you still like to work on this? I can create another PR to work on a builder-based option as well.

alexcrichton:
Addressed in https://github.com/alexcrichton/openssl-src-rs/pull/220

PR #164, opened 2022-10-26
Correctly infer ranlib/ar for cross-gcc toolchains

jonhoo:
A second try at #163.

The GCC convention is that if the toolchain is named `$target-gnu-gcc`, then ranlib and ar will be available as `$target-gnu-gcc-ranlib` and `$target-gnu-gcc-ar` respectively. The code as written would infer them to be `$target-gnu-{ranlib,ar}`, which will only work if the tools from `binutils` (which follow that convention) are on `$PATH`.

I've also updated the logic in line with the `cc` crate to check that the inferred `AR`/`RANLIB` is actually executable before setting it as an override.

See also rust-lang/cc-rs#736.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -346,15 +346,25 @@ impl Build {
             configure.env_remove("CROSS_COMPILE");
 
             // Infer ar/ranlib tools from cross compilers if the it looks like
-            // we're doing something like `foo-gcc` route that to `foo-ranlib`
-            // as well.
+            // we're doing something like `foo-gcc` route that to `foo-gcc-ranlib`
+            // or `foo-ranlib` (for `binutils`) as well.
             if path.ends_with("-gcc") && !target.contains("unknown-linux-musl") {
-                let path = &path[..path.len() - 4];
-                if env::var_os("RANLIB").is_none() {
-                    configure.env("RANLIB", format!("{}-ranlib", path));
-                }
-                if env::var_os("AR").is_none() {
-                    configure.env("AR", format!("{}-ar", path));
+                let suffix = &path[path.len() - 4..];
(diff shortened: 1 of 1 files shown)


jonhoo:
Gentle nudge on this @alexcrichton — this change has now been released in `cc`, so may be good to update this as well so the two crates agree.

alexcrichton:
Sorry but honestly that seems hastily merged with https://github.com/rust-lang/cc-rs/pull/741 following shortly afterwards. I'd prefer to let that bake with user experience before causing more problems here.

jonhoo:
Though after the latest change, this now preserves the existing semantics of preferring the non prefixed binaries. It effectively just adds the gcc-prefixed binaries as another (secondary) alternative to try. The only exception being if the non prefixed binaries don't exist/aren't executable, which is a broken state anyway.

jonhoo:
You mean have `openssl-src` use `cc` to determine the `ar` to use? Yeah, we could probably do that, though I assumed there was a reason it wasn't doing it already? `cc` also doesn't (to my knowledge) detect `ranlib`, so we'd either need to add that to `cc` or retain custom logic here for finding ranlib.

jonhoo:
Gentle nudge on this

jonhoo:
Synced with Alex offline, and I'm going to close this and open a new PR that moves `openssl-src` to just use `cc`'s detection directly. Given that `cc` doesn't yet have ranlib detection, I'll have to land that first though.

jonhoo:
https://github.com/rust-lang/cc-rs/pull/763 is step 1.

jonhoo:
And https://github.com/alexcrichton/openssl-src-rs/pull/173 is the eventual PR that brings that into openssl-src.

Issue #118, opened 2021-12-22
Error while cross compiling from Windows 10 to Raspberry Pi 2 (armv7-unknown-linux-gnueabihf)


adumbidiot:
Hello, I just ran into an issue trying to use the `vendored` feature to cross-compile from Windows 10 to a Raspberry Pi 2. I am using the `nightly-x86_64-pc-windows-msvc` toolchain. The exact error can be seen in the captured stderr when the build fails: 
```
C:/SysGCC/raspberry/bin/arm-linux-gnueabihf-ar: creating apps/libapps.a
  crypto/cversion.c: In function 'OpenSSL_version':
  crypto/cversion.c:38:44: error: incomplete universal character name \U
           return "ENGINESDIR: \"" ENGINESDIR "\"";
                                              ^~~~
  crypto/cversion.c:38:44: warning: unknown escape sequence: '\D'
  crypto/cversion.c:38:44: warning: unknown escape sequence: '\h'
  crypto/cversion.c:38:44: warning: unknown escape sequence: '\d'
  crypto/cversion.c:38:44: warning: unknown escape sequence: '\o'
  crypto/cversion.c:38:44: warning: unknown escape sequence: '\o'
  crypto/cversion.c:38:44: warning: unknown escape sequence: '\o'
  crypto/cversion.c:38:44: warning: unknown escape sequence: '\i'
  make[1]: *** [Makefile:2254: crypto/cversion.o] Error 1
  make: *** [Makefile:177: build_libs] Error 2
``` 

For my build, I specified the following environment variables:
```
PERL = "C:/Users/nathaniel/scoop/apps/msys2/current/usr/bin/perl" 

CC = "C:/SysGCC/raspberry/bin/arm-linux-gnueabihf-gcc"
CXX = "C:/SysGCC/raspberry/bin/arm-linux-gnueabihf-g++"

CMAKE_GENERATOR = "Ninja"
```

I also set the linker through `RUSTFLAGS` as `C:/SysGCC/raspberry/bin/arm-linux-gnueabihf-gcc`. I set my target as `armv7-unknown-linux-gnueabihf`.

I tracked this down to `ENGINESDIR` having backslashes instead of forward slashes, and I tracked it back to the prefix option in the configure script. Currently, this library uses the following to set the prefix:
```rust
if host.contains("pc-windows-gnu") {
    configure.arg(&format!("--prefix={}", sanitize_sh(&install_dir)));
} else {
[...]

PR #165, opened 2022-10-30
impl OPENSSL_SRC_PREFIX_SANTIZE env var

ShootingKing-AM:
- Sanitize path forcefully based on env var
  - 2 use cases as issues #118 
- Dump some helpful build info when build fails


ShootingKing-AM:
Actually I also thought so, but i dont know all kinds of cases people use this crate in. So thought I wouldn't want to make any breaking changes.

PR #173, opened 2022-12-12
Use cc for ar/ranlib detection

jonhoo:
Note that `Command::get_program` and `Command::get_args` both stabilized in Rust 1.57.0, and so implicitly bump this crate's MSRV.

Depends on rust-lang/cc-rs#763.

Replaces #164.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,7 @@
 extern crate cc;
 
 use std::env;
+use std::ffi::OsStr;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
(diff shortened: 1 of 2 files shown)


jonhoo:
Marked this as a draft until the cc-rs PR lands.

jonhoo:
`cargo package` fails because it doesn't respect the `[patch.crates-io]`.

jonhoo:
@alexcrichton The features we needed from `cc` have now landed, so I've marked the PR as ready for review. The biggest discussion point here I think is that this bumps MSRV for the crate since it uses `CommandBuilder::get_args` which requires 1.57.0. We _could_ extend `cc` so that it also provided a mechanism for getting the command and arguments separately, but it feels unnecessary given that we now have the "right" way to do this via `std`.

alexcrichton (approved):

jonhoo:
Do you have a rough sense for when this may land in a release? I have a couple of builds in semi-weird environments that are currently failing and would be fixed by this :)

alexcrichton:
I believe an OpenSSL security release is happening next week, so next week.

jonhoo:
Interesting — that makes me wonder if we might want to release this now to get a heads up in case it breaks anyone's builds. I worry about getting into a position where a security fix goes out at the same time as a change that may break people's builds, as it may mean some people end up not being able to take the security fix.

jonhoo:
Ah, I just realized this only affects the 300 major version. I assume it'd also be okay to backport to 111? I'm happy to do that assuming it's acceptable.

alexcrichton:
Yes that's fine, I can make a release after a PR to the 111 branch

jonhoo:
Back port to 111 in https://github.com/alexcrichton/openssl-src-rs/pull/180

Skepfyr:
Note this does appear to have broken rustup builds:
https://github.com/rust-lang/rustup/issues/3263
https://github.com/sfackler/rust-openssl/issues/1839

alexcrichton:
I don't know how best to fix this myself, unfortunately.

src/lib.rs
```diff
@@ -345,17 +346,25 @@ impl Build {
...
+                // single binary, but instead a multi-argument command like `cmd /c emar.bar`.
+                // We can't convey that through `AR` alone, and so also need to set ARFLAGS.
+                configure.env(
+                    "ARFLAGS",
+                    ar.get_args().collect::<Vec<_>>().join(OsStr::new(" ")),
+                );
+            }
+            let ranlib = cc.get_ranlib();
+            configure.env("RANLIB", ranlib.get_program());
+            if ranlib.get_args().count() == 0 {
```
wangfenjin, line 362: If count is 0, then we set the RANLIBFLAGS to empty string?

Why we need to do that?

PR #184, opened 2023-03-10
Rust version compatibility fix

andrewwhitehead:
It seems that #180 is making use of `[&OsStr]::join` which was not available (broken?) until Rust 1.63: https://github.com/rust-lang/rust/pull/96881/


andrewwhitehead:
I'm not sure why the android builds are failing, unrelated?

amousset:
You can comment out the android tests, the failures are unrelated.

tsxiaofang:
[&OsStr]::join which was not available (broken?) until Rust 1.63, To be compatible with versions before 1.63, join(OsStr:: new ("")) cannot be used, instead with fold(OsString::new(), |mut acc, v|{acc.push(v); acc.push(OsStr::new(" ")); acc})

PR #185, opened 2023-03-18
Fix RANLIB and ARFLAGS

ehuss:
This fixes the use of ARFLAGS and RANLIBFLAGS.

The check for ARFLAGS was inverted, and was only ever giving an empty string for ARFLAGS.

AFAICT, the OpenSSL Makefile does not support RANLIBFLAGS. This updates it so that the flags get stuffed in with RANLIB.


alexcrichton (approved):
Would you like a release for this as well?

ehuss:
Yes, please!

PR #186, opened 2023-03-28
Add support for openssl 3.1

amousset:


alexcrichton:
I don't know much about 3.1.0 vs 3.0.0 myself, but I would naively expect that there wouldn't be a need to maintain a `release/300` branch in the same manner that `release/111` has a branch. Are there breaking changes or otherwise reasons that applications would stay on 3.0.0 vs move over to 3.1.0?

amousset:
> Are there breaking changes or otherwise reasons that applications would stay on 3.0.0 vs move over to 3.1.0?

No breaking changes but the 3.0 branch is an LTS (maintained until 2026-09-07, while 3.1 support ends on 2025-03-14), so it may be interesting for people who want to stick to LTS versions.

sfackler:
OpenSSL switched over to semver in 3.0, so 3.1.x is ABI compatible with 3.0.x unlike 1.0.x and 1.1.x.

sfackler:
Yeah, we should just have a 1.1.0 and 3.x.x track.

alexcrichton:
Ok cool, in that case perhaps this can bump to `300.1.0+....` as a version number? 

Also if you'd like feel free to remove the Android CI builders. I don't know how to fix them off the top of my head and they're otherwise just causing noise.

amousset:
> If not then I'd be tempted to release a version under the 300.*.* track which bumps to 3.1.0 for OpenSSL and go with that. It's already sort of a pain maintaining two branches so I'd personally prefer to not add a third unless there's a reason to

I'm fine with that, updated.

alexcrichton (approved):

Issue #188, opened 2023-04-11
android build fail for missing i686-linux-android-ranlib


tkkcc:
no problem with 111.24
```txt
[[package]]
name = "openssl-src"
version = "111.24.0+1.1.1s"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "3498f259dab01178c6228c6b00dcef0ed2a2d5e20d648c017861227773ea4abd"
dependencies = [
 "cc",
]
```

with 111.25,
```txt
[[package]]
name = "openssl-src"
version = "111.25.2+1.1.1t"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "320708a054ad9b3bf314688b5db87cf4d6683d64cfc835e2337924ae62bf4431"
dependencies = [
 "cc",
]
```
got this error
```txt
/home/bilabila/Android/Sdk/ndk/25.1.8937393/toolchains/llvm/prebuilt/linux-x86_64/bin/llvm-ar: warning: creating libcrypto.a
/bin/sh: line 1: i686-linux-android-ranlib: command not found
/home/bilabila/Android/Sdk/ndk/25.1.8937393/toolchains/llvm/prebuilt/linux-x86_64/bin/llvm-ar: warning: creating libssl.a
/bin/sh: line 1: i686-linux-android-ranlib: command not found
/bin/sh: line 5: i686-linux-android-ranlib: command not found
make: *** [Makefile:313: install_dev] Error 127
thread 'main' panicked at '

```

there is no i686-linux-android-ranlib in ndk 25.2, 25.1 or 24.0

ErikBjare:
Fixed by setting `RANLIB_$target`: https://github.com/ActivityWatch/aw-server-rust/blob/378e8e811fb11110cf3b519ed93001817848061e/compile-android.sh#L54

Where `target` is e.g. `aarch64_linux_android`

likfe:
try my solution https://github.com/alexcrichton/openssl-src-rs/issues/182#issuecomment-1774929052

PR #196, opened 2023-05-12
Sets RANLIB for Android only if using GCC

eycorsican:
Should fix https://github.com/alexcrichton/openssl-src-rs/issues/188

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -358,11 +358,16 @@ impl Build {
                     ar.get_args().collect::<Vec<_>>().join(OsStr::new(" ")),
                 );
             }
-            let ranlib = cc.get_ranlib();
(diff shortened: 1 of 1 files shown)


RadaofTheLight (approved):

elpiel:
There is another issue we've experienced after updating some of our dependencies and applying fix.
We had to add the archiver as well:
https://github.com/alexcrichton/openssl-src-rs/compare/release/111...Stremio:openssl-src-rs:release/111

ErikBjare:
If this PR eliminates the need to set `RANLIB_*`, then it'd be nice to get it merged, so we can strip that out of [our build scripts](https://github.com/ActivityWatch/aw-server-rust/blob/378e8e811fb11110cf3b519ed93001817848061e/compile-android.sh#L54) :)

PR #207, opened 2023-09-11
fix: Sets AR for Android only if using GCC

elpiel:
Fixes an issue with Android builds with ndk version 25 and up (and possibly 24)
Related issue: #188 
Related PR: #196

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -353,22 +353,27 @@ impl Build {
             // prefix, we unset `CROSS_COMPILE` for `./Configure`.
             configure.env_remove("CROSS_COMPILE");
 
-            let ar = cc.get_archiver();
-            configure.env("AR", ar.get_program());
-            if ar.get_args().count() != 0 {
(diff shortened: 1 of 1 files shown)


alexcrichton:
I think this might perhaps be better for the `cc` crate? That it's returning `get_ar` and `get_ranlib` and one of them doesn't work seems like that's best fixed upstream if possible

elpiel:
I'll look into it. I haven't had the opportunity to see what `cc` does and if it's the right fix to go upstream.

elpiel:
@alexcrichton would you consider including this fix as I don't receive any feedback or comments on the issue I created upstream in `cc`?

alexcrichton:
Ah ok, well at least an attempt was made! 

Would you be up for helping out to get Android running on CI though in lieu of upstream support? I've had to disable all the Android CI support historically due to it bitrotting but if more Android-specific support is grown here then I think it would be best to get a CI builder up-and-running to ensure that it keeps working in the future again.

PR #215, opened 2023-10-25
Disable WASI on CI

alexcrichton:
This accidentally worked before but `crypto/uid.c` references nonexistent functions in the WASI SDK which no longer works with more recent versions of Clang in an updated WASI SDK. Currently WASI SDK I think has a bug when compiled against the latest version of LLVM (pulled in with updated rustc) which causes CI failures as well.

I've updated the configuration here to pull in a newer WASI SDK which exposes the failure to build with `crypto/uid.c`, but I unfortunately don't have time to track this further right now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,9 +59,9 @@ jobs:
         - target: x86_64-pc-windows-gnu
           rust: stable
           os: ubuntu-latest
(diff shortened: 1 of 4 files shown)


alexcrichton:
cc @OtaK from https://github.com/alexcrichton/openssl-src-rs/pull/119 as you might be interested in this

PR #216, opened 2023-11-20
Add openembedded targets

mvalenzuelamoixa:


alexcrichton:
Do you have a link to the PR adding these targets in upstream rust-lang/rust?

mvalenzuelamoixa:
This is coming from using the OpenEmbedded rust-common class in bitbake with `TARGET_VENDOR` defaulting to `oe-`: https://git.openembedded.org/openembedded-core/tree/meta/classes-recipe/rust-common.bbclass

I see this is a difference between this version and the one here:
https://github.com/meta-rust/meta-rust/blob/master/classes/rust-common.bbclass

Let me try to understand _why_ that is before going futther.

antoinefaure:
Hi, 
Any update on this @mvalenzuelamoixa ? 
For my architecture I would also need `"aarch64-poky-linux-gnu" => "linux-aarch64"`, could this be added at the same time please ?

PR #217, opened 2023-11-23
Bump to 3.2.0

amousset:
Still working on reducing the size of the crate, a lof of big stuff was added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "openssl-src"
-version = "300.1.6+3.1.4"
+version = "300.2.0+3.2.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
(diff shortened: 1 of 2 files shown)


alexcrichton:
If it's too difficult to trim out the extra bits I can also email crates.io folks to try to get the size limit for this crate increased to 20M for example.

alexcrichton:
Want to try rebasing over #219 and see if this passes CI now?

alexcrichton:
The sizing bits look good (thanks @heiher!) but it looks like there may be new runtime dependencies on Windows? Probably "just" a missing `-l` argument though (unsure which one myself without googling further)

heiher:
> Oh I think that the line you'll need to modify is [around here](https://github.com/alexcrichton/openssl-src-rs/blob/main/src/lib.rs#L690)

Fixed: https://github.com/alexcrichton/openssl-src-rs/pull/221

amousset:
Thanks @heiher, I picked your commit (sorry for the latency, I'm not very available these days).

PR #219, opened 2023-12-04
Exclude some unused docs to reduce the package size

heiher:


alexcrichton:
To confirm, have you tested that this syntax works?

heiher:
> To confirm, have you tested that this syntax works?

Yeah. It works.

PR #237, opened 2024-04-09
Bump openssl to 3.3.0

amousset:


alexcrichton:
I'm not sure how to parse the riscv error precisely, but that looks like we either need to upgrade the assembler/toolchain use to some new syntax or that we should disable the risc-v acceleration and switch it back to linux-generic (or w/e the generic non-asm target is called)

zh-jq-b:
The test docker image is Ubuntu 18.04, how about use version 20.04 or 22.04?

alexcrichton:
Yes usage of 18.04 typically just means it was added in 2018, it's fine to update images.

amousset:
Thanks @zh-jq-b!

PR #239, opened 2024-05-13
fix: build should support lib64 dir for OpenSSL3.x

southorange0929:


alexcrichton:
Thanks! Could you expand a bit more on why this is necessary? For example why hasn't this caused any issues up to now? Is this a recent change upstream?

southorange0929:
For some architectures, OpenSSL version 3.x is installed under the `lib64` directory. I've encountered some dependency query failures when I try to build the x86_64 architecture of openharmony.

This approach is also in sync with [rust-openssl](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/find_normal.rs#L20).

sfackler:
The libdir is overridden to `lib`: https://github.com/alexcrichton/openssl-src-rs/blob/main/src/lib.rs#L191-L192

southorange0929:
You're right, if it ends up being overridden, this PR would be rendered pointless.

PR #244, opened 2024-06-21
use cc_env instead of path to set the CC configure variable

doy-materialize:
this allows wrappers like sccache to work properly


alexcrichton:
Thanks! This repository doesn't have a super high amount of traffic so the CI failures here may or may not be caused by this PR. This seems to me like it should work, but if you're ok with it pushing up a commit that reverts your change, just to see CI run against `main` as-is, will help diagnose if it's specific to this PR or not.

doy-materialize:
okay, this seems to work now.

alexcrichton:
Thanks!

PR #248, opened 2024-09-03
Bump OpenSSL to 3.3.2

alexcrichton:


thalesfragoso:
Hi @alexcrichton. Thanks for your work in this project.

Would you be opposed to mirroring the openssl CVEs to RUSTSEC by tagging this crate ?

alexcrichton:
Seems plausible yeah, but what would that entail? (I'm not sure how to do that myself)

thalesfragoso:
It's basically creating a PR to `RustSec/advisory-db`. More info on [their website](https://rustsec.org/contributing.html).

The idea is to basically copy the official openssl's advisories verbatim. The advantage of doing that is that there are automatic dependency scanners that check rustsec.

The reporter isn't required to be the owner of the crate, so I could also do it when time allows.

alexcrichton:
Ah ok I don't have the time to myself catalog all OpenSSL CVEs and manually mirror them, but if I can do something to help support someone else via this repo that seems reasonable to implement.

PR #250, opened 2024-09-09
Add support for x86_64-pc-windows-gnullvm

janstrohbeck:
In my testing, this was enough to enable support for the [x86_64-pc-windows-gnullvm](https://doc.rust-lang.org/rustc/platform-support/pc-windows-gnullvm.html) target with [llvm-mingw](https://github.com/mstorsjo/llvm-mingw). Without this change, an error is raised because the target is unknown.


alexcrichton:
Thanks! If you'd like and if possible it'd be great to get this tested on CI as well. Not required though

PR #251, opened 2024-09-11
Default enable DSO API to support FIPS module

h0tw1r3:
Unless I'm mistaken, DSO support has no correlation with building shared libraries.

Without DSO support:
* engine support is automatically disabled(according to the docs).
* FIPS modules are not loadable, which typically come from validated compiled versions or the OS vendor.


alexcrichton (approved):
Thanks!

PR #252, opened 2024-09-12
Add aarch64-unknown-openbsd to lib.rs

4n3w:
Required to compile on openbsd -current with a snapdragon X elite cpu


alexcrichton:
Thanks!

PR #253, opened 2024-09-17
Surface the failing command's name from the build error dump

kornelski:
Because [Cargo's formatting of build errors is very noisy](https://github.com/rust-lang/cargo/issues/10159) users struggle to understand what is causing the build to fail.

Here's a [recent case where the failure reason was simple, but the user could not see it](https://users.rust-lang.org/t/in-alpine-rust-can-not-link-ssl-library/117576/5) through all the noise.

----

```diff
    Compiling openssl-sys v0.9.103 (~/openssl-sys/openssl-sys)
+The following warnings were emitted during compilation:
+
+warning: openssl-sys@0.9.103: configuring OpenSSL build: Command 'perl' not found. Is perl installed?
+
 error: failed to run custom build command for `openssl-sys v0.9.103 (~/openssl-sys/openssl-sys)`
 note: To improve backtraces for build dependencies, set the CARGO_PROFILE_DEV_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.
 
 Caused by:
   process didn't exit successfully: `~/openssl-sys/target/debug/build/openssl-sys-60b6752181d583d2/build-script-main` (exit status: 101)
 
 [...]

   --- stderr
-   thread 'main' panicked at ~/openssl-src-rs/src/lib.rs:578:14:
+   thread 'main' panicked at ~/openssl-sys/build/find_vendored.rs:18:39:
 
 
 
   Error configuring OpenSSL build:
-      Command: cd "~/openssl-sys/target/debug/build/openssl-sys-8f90d3753aa6037e/out/openssl-build/build/src" && env -u CROSS_COMPILE AR="ar" CC="cc" RANLIB="ranlib" "perl" "./Configure" "--prefix=~/openssl-sys/target/debug/build/openssl-sys-8f90d3753aa6037e/out/openssl-build/install" "--openssldir=/usr/local/ssl" "no-dso" "no-shared" "no-ssl3" "no-tests" "no-comp" "no-zlib" "no-zlib-dynamic" "--libdir=lib" "no-md2" "no-rc5" "no-weak-ssl-ciphers" "no-camellia" "no-idea" "no-seed" "darwin64-arm64-cc" "-nologo" "-MD" "-O2" "-Z7" "-Brepro" "-mmacosx-version-min=11.0"
-      Failed to execute: No such file or directory (os error 2)
+      Command 'perl' not found. Is perl installed?
+      Command failed: cd "~/openssl-sys/target/debug/build/openssl-sys-3ae48db4843d8fdf/out/openssl-build/build/src" && env -u CROSS_COMPILE AR="ar" CC="cc" RANLIB="ranlib" "perl" "./Configure" "--prefix=~/openssl-sys/target/debug/build/openssl-sys-3ae48db4843d8fdf/out/openssl-build/install" "--openssldir=/usr/local/ssl" "no-shared" "no-ssl3" "no-tests" "no-comp" "no-zlib" "no-zlib-dynamic" "--libdir=lib" "no-md2" "no-rc5" "no-weak-ssl-ciphers" "no-camellia" "no-idea" "no-seed" "darwin64-arm64-cc" "-nologo" "-MD" "-O2" "-Z7" "-Brepro" "-mmacosx-version-min=11.0"
```

* Cleverness of the `Debug` impl of `Command` obscures the command name that has been run. This change prints the command name alone.
* NotFound error is handled with a dedicated message.
* Uses `cargo:warning` to also print the message above the stdout and stderr dumps.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,45 +60,47 @@ impl Build {
         self
     }
 
-    fn cmd_make(&self) -> Command {
-        let host = &self.host.as_ref().expect("HOST dir not set")[..];
-        if host.contains("dragonfly")
-            || host.contains("freebsd")
-            || host.contains("openbsd")
-            || host.contains("solaris")
-            || host.contains("illumos")
-        {
-            Command::new("gmake")
-        } else {
-            Command::new("make")
-        }
+    fn cmd_make(&self) -> Result<Command, &'static str> {
+        let host = &self.host.as_ref().ok_or("HOST dir not set")?[..];
+        Ok(
+            if host.contains("dragonfly")
+                || host.contains("freebsd")
+                || host.contains("openbsd")
+                || host.contains("solaris")
+                || host.contains("illumos")
+            {
+                Command::new("gmake")
+            } else {
+                Command::new("make")
+            },
+        )
     }
 
     #[cfg(windows)]
-    #[track_caller]
     fn check_env_var(&self, var_name: &str) -> Option<bool> {
-        env::var_os(var_name).map(|s| {
+        env::var_os(var_name).and_then(|s| {
             if s == "1" {
                 // a message to stdout, let user know asm is force enabled
                 println!(
-                    "{}: nasm.exe is force enabled by the \
+                    "cargo:warning={}: nasm.exe is force enabled by the \
                     'OPENSSL_RUST_USE_NASM' env var.",
                     env!("CARGO_PKG_NAME")
                 );
-                true
+                Some(true)
             } else if s == "0" {
                 // a message to stdout, let user know asm is force disabled
                 println!(
-                    "{}: nasm.exe is force disabled by the \
+                    "cargo:warning={}: nasm.exe is force disabled by the \
                     'OPENSSL_RUST_USE_NASM' env var.",
                     env!("CARGO_PKG_NAME")
                 );
-                false
+                Some(false)
             } else {
-                panic!(
-                    "The environment variable {} is set to an unacceptable value: {:?}",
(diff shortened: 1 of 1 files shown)


kornelski:
I've also added `try_build`, so that I can make `openssl-sys` report the error without a noisy backtrace.

bjorn3:
Instead of using panic, maybe using eprintln!() + std::process::exit(1) would work? openssl-src is probably not called in any context where recovering from an unexpected build error is necessary, and you added try_build for cases where build errors are expected to happen.

kornelski:
I've updated it with a clean exit, and converted other panics into errors.

src/lib.rs
```diff
@@ -121,23 +124,35 @@ impl Build {
...
     pub fn build(&mut self) -> Artifacts {
-        let target = &self.target.as_ref().expect("TARGET dir not set")[..];
-        let host = &self.host.as_ref().expect("HOST dir not set")[..];
-        let out_dir = self.out_dir.as_ref().expect("OUT_DIR not set");
+        match self.try_build() {
+            Ok(a) => a,
+            Err(e) => {
+                println!("cargo:warning=openssl-src: failed to build OpenSSL from source");
+                eprintln!("\n\n\n{e}\n\n\n");
+                std::process::exit(101); // same as a panic
```
bjorn3, line 134: It should probably use 1 as exit code given that it isn't a bug in the build script that caused a crash.

kornelski: ok

alexcrichton:
Thanks for this!

PR #255, opened 2024-10-22
Bump OpenSSL to 3.4.0

alexcrichton:
Updating to the most recent release.

PR #256, opened 2024-10-22
Bump OpenSSL to 3.4.0

weihanglo:
This could potentially fix <https://github.com/sfackler/rust-openssl/issues/2163> though I haven't tested it.

See also <https://github.com/rust-lang/cargo/issues/13546#issuecomment-2429345001>.

Any manual test we could do when bumping submodule?


weihanglo:
You bit me :) https://github.com/alexcrichton/openssl-src-rs/pull/255

PR #259, opened 2024-11-16
Set no-module to fix legacy provider

### LINKED ISSUE #257: 3.4.0 release breaks the legacy provider
We ran into this error over when updating OpenSSL in Vector: [vectordotdev/vector#21587](https://github.com/vectordotdev/vector/pull/21587)

```
thread 'tls::settings::test::from_options_pkcs12' panicked at lib/vector-core/src/tls/settings.rs:682:85:
called `Result::unwrap()` on an `Err` value: ErrorStack([Error { code: 310378599, library: "DSO support routines", function: "dlfcn_load", reason: "could not load the shared library", file: "crypto/dso/dso_dlfcn.c", line: 118, data: "filename(/home/runner/work/vector/vector/target/debug/build/openssl-sys-7970cd031f6a569d/out/openssl-build/install/lib/ossl-modules/legacy.so): /home/runner/work/vector/vector/target/debug/build/openssl-sys-7970cd031f6a569d/out/openssl-build/install/lib/ossl-modules/legacy.so: cannot open shared object file: No such file or directory" }, Error { code: 310378599, library: "DSO support routines", function: "DSO_load", reason: "could not load the shared library", file: "crypto/dso/dso_lib.c", line: 147 }, Error { code: 126353445, library: "common libcrypto routines", function: "provider_init", file: "crypto/provider_core.c", line: 950, data: "name=legacy" }])
stack backtrace:
   0:     0x564eac5a8cd5 - std::backtrace_rs::backtrace::libunwind::trace::h649ab3318d3445c5
```

Ref: https://github.com/alexcrichton/openssl-src-rs/pull/255#issuecomment-2441341574

sfackler:
I'm honestly pretty confused as to how this was working previously - unless no-module is set, you'll end up with a legacy provider dynamic object that needs to exist in the openssldir: https://github.com/openssl/openssl/issues/17679

Tested here: https://github.com/sfackler/rust-openssl/pull/2331

Closes #257

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -197,6 +197,7 @@ impl Build {
         configure
             // No shared objects, we just want static libraries
             .arg("no-shared")
+            .arg("no-module")
             // Should be off by default on OpenSSL 1.1.0, but let's be extra sure
             .arg("no-ssl3")
             // No need to build tests, we won't run them anyway


alexcrichton (approved):
Thanks!

PR #263, opened 2025-04-08
Bump to OpenSSL 3.5.0

alexcrichton:


alexcrichton:
I've requested an increase on the crate upload for this crate from 10M to 15M and crates.io has graciously granted this. If anyone knows how to remove even more from the tarball to cut the size down below 10M I think that would be good to add regardless.

PR #271, opened 2025-10-13
Add support for sparc64-unknown-linux-gnu.

thejpster:
OpenSSL supports Linux on 64-bit SPARC.

Add the missing entry so we can map `sparc64-unknown-linux-gnu` to the correct OpenSSL build configuration.

With this fix, I was able to build openssl-sys in vendored mode for the `sparc64-unknown-linux-gnu` target.


thejpster:
Done - #274

PR #283, opened 2026-04-18
Fix Apple simulator targets and add visionOS support

cypherair:
This updates Apple mobile target handling in `openssl-src` in two steps.

First, it maps iOS simulator Rust targets to OpenSSL's architecture-specific simulator configurations:

- `x86_64-apple-ios` -> `iossimulator-x86_64-xcrun`
- `aarch64-apple-ios-sim` -> `iossimulator-arm64-xcrun`

OpenSSL already provides these configure targets. Using the generic `iossimulator-xcrun` can still produce a simulator archive for the requested architecture, but the generated OpenSSL configuration falls back to generic simulator settings such as `THIRTY_TWO_BIT` and `OPENSSL_NO_ASM`. The architecture-specific targets produce the expected 64-bit configuration and enable the matching assembly settings.

Second, this adds support for Rust's visionOS targets:

- `aarch64-apple-visionos` -> `ios64-cross`
- `aarch64-apple-visionos-sim` -> `iossimulator-arm64-xcrun`

For visionOS, the existing Apple mobile `-isysroot` handling is extended to `apple-visionos`, and the compiler override selects the corresponding SDK:

- device: `xcrun -sdk xros cc`
- simulator: `xcrun -sdk xrsimulator cc`

This keeps using OpenSSL's existing Apple configure targets and does not add new OpenSSL platform configuration files.

Tested locally with:

```bash
cargo test --manifest-path testcrate/Cargo.toml --target x86_64-apple-ios --no-run -vv
cargo test --manifest-path testcrate/Cargo.toml --target aarch64-apple-ios-sim --no-run -vv
cargo test --manifest-path testcrate/Cargo.toml --target aarch64-apple-visionos --no-run -vv
cargo test --manifest-path testcrate/Cargo.toml --target aarch64-apple-visionos-sim --no-run -vv
```

I also checked the generated `openssl/configuration.h` for these targets and confirmed they use `SIXTY_FOUR_BIT_LONG`, leave `THIRTY_TWO_BIT` undefined, and no longer define `OPENSSL_NO_ASM`.


alexcrichton (approved):
Thanks!

